PR #17, opened 2014-09-22
Changed to safely call  SWF functions for disposing videojs

t2y:
The media-source plugin calls SWF functions and I got an error when I have called videojs.dispose() as below. I confirmed with videojs-4.8.2, videojs-swf-4.4.3, videojs-contrilb-hls-plugin-0.9.0 and videojs-contrib-media-source-0.3.0 on IE10, IE11 and Chrome35, but it didn't occurred on Firefox31 somehow. I think it should be call after checking those function's existence.

```js
TypeError: undefined is not a function
    at abort (http://path/to/videojs-media-sources.js:217:26)
    at videojs.Hls.dispose (http://path/to/videojs.hls.js:247:23)
    at vjs.Player.dispose (http://path/to/video.dev.js:3448:30)
    ...
```

### DIFF
--- a/src/videojs-media-sources.js
+++ b/src/videojs-media-sources.js
@@ -187,10 +187,12 @@
 
           // bypass normal ExternalInterface calls and pass xml directly
           // EI can be slow by default
-          self.source.swfObj.CallFunction('<invoke name="vjs_appendBuffer"' +
-                                          'returntype="javascript"><arguments><string>' +
-                                          b64str +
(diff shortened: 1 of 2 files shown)


dmlap:
@t2y I'm not able to reproduce this in Chrome with contrib-mse 0-3.0 and contrib-hls 0.10. Is there a live example we could look at somewhere to debug it? Please re-open this issue if you're still seeing it.

PR #20, opened 2015-01-07
Use setTimeout instead of requestAnimationFrame

dmlap:
#15 changed back to just use setTimeout(). Swapping between requestAnimationFrame() and setTimeout() is nice but I don't think the benefit it adds is worth the additional code complexity. Added travis support and fixed up tests in phantomjs.

### DIFF
--- a/src/videojs-media-sources.js
+++ b/src/videojs-media-sources.js
@@ -162,7 +162,7 @@
             scheduleTick(append);
           }
         },
-        append = function() {4
+        append = function() {
           var chunk, i, length, payload, maxSize,
               binary = '';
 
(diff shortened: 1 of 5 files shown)


src/videojs-media-sources.js
```diff
@@ -144,8 +145,25 @@
 
         // the total number of queued bytes
         bufferSize = 0,
+        scheduleTick = function(func) {
```
gkatsev: we're still swapping between the two here?

gkatsev:
Seems like we're still swapping between rAF and setTimeout in `scheduleTick` which conflicts with your comment here unless there's more code changes to come?

src/videojs-media-sources.js
dmlap: Forgot a commit...

gkatsev:
LGTM

PR #22, opened 2015-04-01
Browserify friendly

stacy-andrews:
We are trying to use the media sources within a browserify-based application.

We've had to make minor modifications to support this.


forbesjo:
Sorry for the delay but could you rebase this? Thank you.

PR #25, opened 2015-05-14
Throttle updating

dmlap:
HTML5 SourceBuffers do not allow appends while an update is in progress. We don't have a coded frame processing loop to run but we do asynchronously shuttle bytes into the object tag. Throw an error if updates are requested during that period. Expose the updating property on SourceBuffers for convenient querying.


forbesjo:
LGTM

Issue #28, opened 2015-07-09
b64str should be declared


khejing:
I use videojs-contrib-media-sources with webpack, and compile it with Babel. Babel use strict mode by default, so videojs-media-sources.js will throw an exception in browser, just because b64str has not been declared. For now I can use imports-loader to declare b64str in webpack config file, but I think it should be fixed, right?

gkatsev:
Oops. It should definitely be declared. Want to submit a PR for it?

PR #29, opened 2015-07-10
add b64str declaration

khejing:
fix https://github.com/videojs/videojs-contrib-media-sources/issues/28


forbesjo:
Thank you for your contribution but I will be closing this because it has been fixed in a previous change.

PR #32, opened 2015-09-01
Make Transmuxer WebWorker and Defer creation of SourceBuffers

imbcmdth:
* Make a transmuxer WebWorker.
* Include a build system to inline the WebWorker + `importScripts` into a blob.
* Defer the creation of `SourceBuffers` until after we start getting segments back from the transmuxer so that we know what SourceBuffers need to be created (Audio-only, Video-only, or both.)
* Fix tests!


src/videojs-media-sources.js
```diff
@@ -100,64 +98,156 @@
...
-        if (event.type === 'video') {
-          buffer = this.videoBuffer_;
-          this.videoUpdating_ = false;
-        } else {
-          buffer = this.audioBuffer_;
-          this.audioUpdating_ = false;
-        }
-        if (this.timestampOffset !== undefined) {
-          buffer.timestampOffset = this.timestampOffset;
+      this.transmuxer_ = new Worker(videojs.MediaSource.webWorkerURI || '/src/transmuxer_worker.js');
```
dmlap: Decided not to get fancy, huh?

src/videojs-media-sources.js
```diff
@@ -100,64 +98,156 @@
...
+            buffer,
+            segment = event.data;
+          // Cast to type
+          segment.data = new Uint8Array(segment.data);
+
+          // If any sourceBuffers have not been created, do so now
+          if (segment.type === 'video') {
+            if (!self.videoBuffer_) {
+              // 4d400d
+              // 42c01e
```
dmlap: Leftovers?

src/videojs-media-sources.js
```diff
@@ -100,64 +98,156 @@
...
+              (this.audioBuffer_ && this.audioBuffer_.buffered.length === 0)) {
             return videojs.createTimeRange();
           }
-          start = Math.max(this.videoBuffer_.buffered.start(0),
-                           this.audioBuffer_.buffered.start(0));
-          end = Math.min(this.videoBuffer_.buffered.end(0),
-                         this.audioBuffer_.buffered.end(0));
+
+          // TODO: Calculate and return the true intersection of the buffered
+          //       property
```
dmlap: What remains to be done here?

src/videojs-media-sources.js
```diff
@@ -100,64 +98,156 @@
...
+    processPendingSegments_: function() {
+      var segment, buffer, i, offset,
+        videoSegments = [],
+        audioSegments = [],
+        videoBytes = 0,
+        audioBytes = 0;
+
+      // Sort segments into separate video/audio arrays and
+      // keep track of their total byte lengths
+      while ((segment = this.pendingBuffers_.shift())) {
```
dmlap: This would be fantastic C code but I feel it's a bad practice in javascript.

src/videojs-media-sources.js
```diff
@@ -100,64 +98,156 @@
...
+      if (audioBytes) {
+        this.concatAndAppendSegments_(audioSegments,audioBytes, this.audioBuffer_);
+      }
+
+      // We are no longer in the internal "updating" state
+      this.bufferUpdating_ = false;
+    },
+    concatAndAppendSegments_: function(segments, totalBytes, destinationBuffer) {
+      var offset = 0;
+      var tempBuffer = new Uint8Array(totalBytes);
```
dmlap: Multiple `var`s again. I would love if we could just babelify this project so I can stop nagging about it :)

dmlap:
A couple tiny nitpicks, otherwise LGTM

src/videojs-media-sources.js
imbcmdth: Well, for one, we only return a TimeRanges object with a single span even if more than one span exists. I'd like to think that for completeness we should consider returning a real object that is the intersection of both audio and video.

Although, if unification of the two streams is successful this whole "TODO" disappears. So.. yay!

src/videojs-media-sources.js
imbcmdth: I do it automatically. I can't help it. I'm a monster!

src/videojs-media-sources.js
imbcmdth: That's it. A reduce it is!

src/videojs-media-sources.js
```diff
@@ -100,64 +98,166 @@
...
+          },
+          audio: {
+            segments: [],
+            bytes: 0
+          }
+        };
+
+      // Sort segments into separate video/audio arrays and
+      // keep track of their total byte lengths
+      sortedSegments = this.pendingBuffers_.reduce(function (segmentObj, segment) {
```
forbesjo: Reduce doesn't seem to be available in IE8, maybe you can add a babel step if you want the code to be cleaner

PR #33, opened 2015-09-08
Cross streams

imbcmdth:
Add support for a 'combined' segment type that has both audio and video tracks.


src/videojs-media-sources.js
```diff
@@ -132,6 +132,17 @@
               self.audioBuffer_.addEventListener('updateend',
                                                  aggregateUpdateHandler(self, 'videoBuffer_', 'updateend'));
             }
+          } else if (segment.type === 'combined') {
```
dmlap: This would be a significant change but I wonder if we should just use MIME types for `segment.type` across the board. Slightly harder to read but more consistent.

PR #35, opened 2015-09-09
Translate in-band captions to webvtt cues

dmlap:
Pick up caption arrays on transmuxed segments and convert them into WebVTT cues on a programmatically generated captions track.

### DIFF
--- a/src/videojs-media-sources.js
+++ b/src/videojs-media-sources.js
@@ -280,7 +280,6 @@
       this.pendingBuffers_.length = 0;
 
       // We are no longer in the internal "updating" state
-      this.pendingBuffers_ = [];
       this.bufferUpdating_ = false;
     },
     /**
--- a/src/videojs-media-sources.js
+++ b/src/videojs-media-sources.js
@@ -37,6 +37,18 @@
           this.settings_.mode === 'html5') {
         self = new window.MediaSource();
         interceptBufferCreation(self);
+
+        // capture the associated player when the MediaSource is
+        // successfully attached
+        self.addEventListener('sourceopen', function() {
+          var video = document.querySelector('[src="' + self.url_ + '"]');
+
+          if (!video && video.parentNode) {
+            return;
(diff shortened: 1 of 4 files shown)


src/videojs-media-sources.js
```diff
@@ -247,6 +280,7 @@
       this.pendingBuffers_.length = 0;
 
       // We are no longer in the internal "updating" state
+      this.pendingBuffers_ = [];
```
imbcmdth: No longer needed because I set pendingBuffers_ length to zero above! :)

src/videojs-media-sources.js
```diff
@@ -214,10 +235,12 @@
       var sortedSegments = {
           video: {
             segments: [],
+            captions: [],
             bytes: 0
           },
           audio: {
             segments: [],
+            captions: [],
```
imbcmdth: I though only video can have captions?

src/videojs-media-sources.js
```diff
@@ -37,6 +37,18 @@
           this.settings_.mode === 'html5') {
         self = new window.MediaSource();
         interceptBufferCreation(self);
+
+        // capture the associated player when the MediaSource is
+        // successfully attached
+        self.addEventListener('sourceopen', function() {
+          var video = document.querySelector('[src="' + self.url_ + '"]');
```
imbcmdth: This was a good call by @gkatsev!

imbcmdth:
Minor nits. Otherwise, LGTM

src/videojs-media-sources.js
dmlap: Incredibly good call.

gkatsev: lol, so bad.

src/videojs-media-sources.js
dmlap: That's true in practice but I thought this simplified the cases downstream.

PR #36, opened 2015-09-10
Add a watch system to automatically build distribution files

imbcmdth:
Clean up and add a watch task to rebuilt the distribution files when a change is detected in this project or in the main dependency: mux.js.


dmlap:
LGTM.

PR #37, opened 2015-09-10
Allow codecs to be passed into source buffer creation

dmlap:
Some codecs (HE-AAC audio) are hard to detect ahead of time and can cause browsers to fail if they're not properly specified. Allow downstream developers to override the default codec selection if they have out-of-band information about the codecs their content uses.


imbcmdth:
LGTM :-1:

PR #39, opened 2015-09-17
Mse discontinuity

imbcmdth:
To support discontinuity we leverage the `timestampOffset` feature of MSE. Whenever a `timestampOffset` is set on our polyfill, (for example by contrib-hls) we force the transmuxer to reset the *BaseMediaDecodeTime* calculations so that the next segment will start at 0. When we receive that segment back from the transmuxer, we then set the `timestampOffset` of the real native source buffers to the value we were given and start appending segments.

### DIFF
--- a/src/transmuxer_worker.js
+++ b/src/transmuxer_worker.js
@@ -51,7 +51,7 @@ var wireTransmuxerEvents = function (transmuxer) {
 };
 
 /**
- * All incoming messages route through this hash. If not function exists
+ * All incoming messages route through this hash. If no function exists
  * to handle an incoming message, then we ignore the message.
  */
 var messageHandlers = {
@@ -70,10 +70,8 @@ var messageHandlers = {
    * default options if `init` was never explicitly called
    */
   defaultInit: function () {
-    if (!transmuxer) {
-      transmuxer = new muxjs.mp2t.Transmuxer();
-      wireTransmuxerEvents(transmuxer);
-    }
+    transmuxer = new muxjs.mp2t.Transmuxer();
+    wireTransmuxerEvents(transmuxer);
   },
   /**
    * push
@@ -81,8 +79,6 @@ var messageHandlers = {
    * processing
    */
   push: function (data) {
-    this.defaultInit();
-
     // Cast array buffer to correct type for transmuxer
     var segment = new Uint8Array(data.data);
     transmuxer.push(segment);
@@ -93,8 +89,6 @@ var messageHandlers = {
    * begin at a baseMediaDecodeTime of 0
    */
   resetBaseMediaDecodeTime: function (data) {
-    this.defaultInit();
-
     transmuxer.resetBaseMediaDecodeTime();
   },
   /**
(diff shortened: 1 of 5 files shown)


src/transmuxer_worker.js
```diff
@@ -17,39 +17,100 @@ importScripts('../node_modules/mux.js/lib/metadata-stream.js');
...
+    });
+  }
+
+  transmuxer.on('done', function (data) {
+    postMessage({ action: 'done' });
+  });
+};
 
+/**
+ * All incoming messages route through this hash. If not function exists
```
dmlap: > If not function exists

src/transmuxer_worker.js
```diff
@@ -17,39 +17,100 @@ importScripts('../node_modules/mux.js/lib/metadata-stream.js');
...
+    transmuxer = new muxjs.mp2t.Transmuxer(data && data.options);
+    wireTransmuxerEvents(transmuxer);
+  },
+  /**
+   * defaultInit
+   * Is called before every function and initializes the transmuxer with
+   * default options if `init` was never explicitly called
+   */
+  defaultInit: function () {
+    if (!transmuxer) {
```
dmlap: What if this was just moved into `onmessage` below? Then we wouldn't have to remember to call it in every future message handler we create.

src/videojs-media-sources.js
```diff
@@ -301,9 +365,9 @@
           offset += segment.byteLength;
         });
 
-        // Set timestampOffset if we have been given one
-        if (this.timestampOffset !== undefined) {
-          destinationBuffer.timestampOffset = this.timestampOffset;
+        // Set timestampOffset if it was changed before this append
```
dmlap: Could we just set the timestampOffset on the destination buffer immediately when the property setter is invoked?

dmlap:
We need a test to ensure that `vjs_discontinuity` is invoked on the Flash source buffer when timestampOffset is modified.

imbcmdth:
We call vjs_discontinuity at https://github.com/videojs/videojs-contrib-media-sources/pull/39/files#diff-2e6098d381c8003336dd711b692da714R537

src/videojs-media-sources.js
imbcmdth: Maybe.

The benefit to doing it this way is that we don't have to worry about the buffers existing when `timestampOffset` is set - the offset is set only when we have done everything we needed to do to create the sourcebuffers but right before the actual `append`.

The real source buffers can be created long after the "virtual" source buffer is created. You may want to create a Media-Source object and set the `timestampOffset` before you ever start pumping data into the transmuxer (and therefore create the source buffers). This allows for that contingency without introducing too many complications.

dmlap: Ugh. I worry this bit of asynchronicity is going to confuse us down the road. What about if we set timestampOffset directly on any buffers that are already created and make sure to check and apply the current timestamp offset as part of our lazy buffer creation process?

imbcmdth: I think that introduces the potential for more confusion and complicates reasoning about the code.

The way it is right now, `timestampOffset` is set in exactly *one* place and that place is right before we `appendBuffer` which is exactly how a user would use `timestampOffset` in normal MSE code (ie. the `timestampOffset` is pretty much set exactly where one would expect it to be set.)

dmlap: `timestampOffset` is side-effecting and you're shifting that side-effect in time. I don't think I'm going out on a limb to say that lazy IO is confusing. Assuming some amount of laziness is required, my proposal:

- makes the "normal" case strict
- gets rid of an extra piece of state to manage (or screw up)

dmlap: We decided to toss the transmuxer and recreate it whenever we need to [reset the base media decode time](https://github.com/imbcmdth/videojs-contrib-media-sources/commit/1eb75f409a1ce12d25030c8a052244dff5fd7130).

PR #40, opened 2015-09-22
Update Flash handling for later vjs5 RCs

dmlap:
player.tech has been renamed to player.tech_ and made "private". Use a reference on the SWF object instead so we don't break any rules.


imbcmdth:
LGTM

PR #44, opened 2015-09-28
Added loadstart to flash side since we need to emit it ourselves

imbcmdth:
Related to https://github.com/videojs/videojs-contrib-hls/pull/400


saxena-gaurav:
LGTM

dmlap:
I think this is too late in the loading process, for the reasons mentioned in https://github.com/videojs/videojs-contrib-hls/issues/400

PR #46, opened 2015-09-29
Remove cues for both captions and id3 tags when we call `remove` to clear a buffered region

imbcmdth:
Extends the remove function to also delete any cues that are overlapping the region so that we don't end up with duplicate cues.

### DIFF
--- a/test/media-sources_test.js
+++ b/test/media-sources_test.js
@@ -175,6 +175,70 @@
     equal(messages[0].action, 'resetTransmuxer', 'reset called on transmuxer');
   });
 
+  test('calling remove deletes cues and invokes remove on any extant source buffers', function(){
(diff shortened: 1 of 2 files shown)


test/media-sources_test.js
```diff
@@ -175,6 +175,70 @@
     equal(messages[0].action, 'resetTransmuxer', 'reset called on transmuxer');
   });
 
+  test('remove on the fake source buffer calls remove on the real ones and removes cues', function(){
```
dmlap: Some punctuation in the description maybe?

dmlap:
I'd love to get more of the skipped tests in the dev branch of contrib-hls ported over here but this is a good first step. LGTM

PR #48, opened 2015-10-07
Less legacy

imbcmdth:
Changes to both the flash side and the mse side to support the massive refactoring in: https://github.com/videojs/mux.js/pull/27


src/videojs-media-sources.js
```diff
@@ -724,6 +725,53 @@ deprecateOldCue = function(cue) {
       this.trigger({ type: 'updateend' });
     },
 
+    receiveBuffer_: function(segment) {
+      // create an in-band caption track if one is present in the segment
+      if (segment.captions &&
```
dmlap: This looks like copy-paste of the HTML in-band captions and metadata translation, yes? If so, we should figure out a way to share more of that logic.

dmlap:
lgtm

PR #50, opened 2015-10-21
Add appendWindowStart to source buffers

dmlap:
Chrome 45 has issues with appending over previously buffered areas. Expose appendWindowStart so we can avoid overlays for now.


dmlap:
Reviewed offline by @imbcmdth

PR #51, opened 2015-10-30
Parse the codecs string into separate entries so that we can create split audio/video buffers

imbcmdth:
These are the minimal set of changes required to support split audio and video source buffers.

### DIFF
--- a/src/videojs-media-sources.js
+++ b/src/videojs-media-sources.js
@@ -172,13 +172,18 @@ addTextTrackData = function (sourceHandler, captionArray, metadataArray) {
   };
 
   addSourceBuffer = function(type) {
-    var audio, video, buffer, codecs;
+    var
+      buffer,
+      codecs,
+      avcCodec,
+      mp4aCodec,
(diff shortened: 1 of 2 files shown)


gkatsev:
The tests are failing because of spaces in the codec output. The expects have them and actuals don't.

src/videojs-media-sources.js
```diff
@@ -216,10 +236,12 @@ addTextTrackData = function (sourceHandler, captionArray, metadataArray) {
       // append muxed segments to their respective native buffers as
       // soon as they are available
       this.transmuxer_ = new Worker(videojs.MediaSource.webWorkerURI || '/src/transmuxer_worker.js');
+      this.transmuxer_.postMessage({action:'init', options: {remux: false}});
```
imbcmdth: This is *technically* the only thing that is required to get non-combined output!

dmlap: Amazing!

dmlap:
Beauty.

PR #52, opened 2015-10-30
remove extraneous condition

saxena-gaurav:


dmlap:
How do you know this condition was extraneous?

misteroneill:
@dmlap Because `video.parentNode` can't possibly be truthy if `video` is falsy?

dmlap:
But video can be truthy without it having a parent node (if, for instance, the video element has been created but not attached to the document anywhere yet). I can't recall specific circumstances where that would occur but it seems weird to me that someone would have added that check without good reason.

/cc @saxena-gaurav

misteroneill:
I considered that at first (it may have been intended to be `if (!video || !video.parentNode) {`, but `document.querySelector` won't return a node that's not in the `document` (just attempted this).

dmlap:
@misteroneill good point there... alright, I can't think of a reason for the check either. Thanks for going over it with me.

PR #53, opened 2015-11-02
More intelligently cull tags before pushing them into Flash

imbcmdth:
Always cull tags that have a PTS before the end of buffer in addition to those that have a PTS before the currentTime


dmlap:
*Probably* needs a test, otherwise LGTM

PR #54, opened 2015-11-05
Flash fixes

imbcmdth:
Some changes to make the Flash MSE polyfill act more like native MSE. 
* `vjs_endOfStream` now called directly from `endOfStream`
* setting a `duration` will call `remove` (which eventually calls 'updateend') if the new duration is less than the old duration
* properly filter tags based on pts since tags are in dts order and the first pts greater or equal to the `targetPts` may not signal the end of filtering

### DIFF
--- a/src/videojs-media-sources.js
+++ b/src/videojs-media-sources.js
@@ -614,7 +614,20 @@ addTextTrackData = function (sourceHandler, captionArray, metadataArray) {
         return this.swfObj.vjs_getProperty('duration');
       },
       set: function(value){
+        var
+          i,
+          oldDuration = this.swfObj.vjs_getProperty('duration');
+
         this.swfObj.vjs_setProperty('duration', value);
+
+        if (value <= oldDuration) {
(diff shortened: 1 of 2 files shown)


src/videojs-media-sources.js
```diff
@@ -614,7 +614,20 @@ addTextTrackData = function (sourceHandler, captionArray, metadataArray) {
         return this.swfObj.vjs_getProperty('duration');
       },
       set: function(value){
+        var
+          i,
+          oldDuration = this.swfObj.vjs_getProperty('duration');
+
         this.swfObj.vjs_setProperty('duration', value);
+
+        if (value <= oldDuration) {
```
dmlap: Why `<=` instead of just `<`?

test/media-sources_test.js
```diff
@@ -1106,15 +1106,20 @@
     mediaSource.swfObj.vjs_endOfStream = function() {
```
dmlap: I think the description of this test should be updated now that we don't actually call `vjs_endOfStream` after the last append.

PR #55, opened 2015-11-15
Implement experimental addSeekableRange()

dmlap:
Create a new method on MediaSources to address https://github.com/w3c/media-source/issues/5. Since we will now have to massage duration and seekable values on the MediaSource itself, wrap the native MediaSource implementation instead of just decorating it. Remove some repeated code in tests.


imbcmdth:
LGTM.

I especially like how you have gotten rid of that strange monkey-patched MediaSource object thingy :smile:

PR #56, opened 2015-11-18
Move timestampOffset into transmuxer (via baseMediaDecodeTime)

imbcmdth:
With this change we never change the underlying sourcebuffer's timestampOffset.

### DIFF
--- a/src/transmuxer_worker.js
+++ b/src/transmuxer_worker.js
@@ -90,28 +90,20 @@ var messageHandlers = {
   /**
    * reset
    * Recreate the transmuxer so that the next segment added via `push`
-   * begins at a baseMediaDecodeTime of 0
+   * start with a fresh transmuxer
    */
-  reset: function (data) {
-    if (data && data.options) {
-      var newOptions = {}, key;
-
-      for (key in initOptions) {
-        if (initOptions.hasOwnProperty(key)) {
-          newOptions[key] = initOptions[key];
-        }
-      }
-
-      for (key in data.options) {
-        if (data.options.hasOwnProperty(key)) {
-          newOptions[key] = data.options[key];
-        }
-      }
-      // recreate the transmuxer
-      this.init({options: newOptions});
-    } else {
-      this.defaultInit();
-    }
+  reset: function () {
+    this.defaultInit();
+  },
+  /**
+   * setTimestampOffset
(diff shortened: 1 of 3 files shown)


src/videojs-media-sources.js
```diff
@@ -318,16 +318,14 @@ addTextTrackData = function (sourceHandler, captionArray, metadataArray) {
...
 
-            if (this.videoBuffer_) {
-              this.videoBuffer_.timestampOffset = val;
-            }
-            if (this.audioBuffer_) {
-              this.audioBuffer_.timestampOffset = val;
-            }
-
             // We have to tell the transmuxer to reset the baseMediaDecodeTime to
             // zero for the next segment
```
dmlap: We don't always reset to zero now, this comment should be updated.

dmlap:
Relies on videojs/mux.js#39 and there are still unresolved questions in that PR.

src/transmuxer_worker.js
```diff
@@ -88,15 +88,24 @@ var messageHandlers = {
...
   },
   /**
+   * setTimestampOffset
+   * Set the value that will be used as the `baseMediaDecodeTime` time for the
+   * next segment pushed in. Subsequent segments will have their `baseMediaDecodeTime`
+   * set relative to the first based on the PTS values.
+   */
+  setTimestampOffset: function (data) {
+    var timestampOffset = data.timestampOffset || 0;
+    transmuxer.setBaseMediaDecodeTime(Math.round(timestampOffset * 90000));
```
dmlap: Why round?

dmlap: Looked again, I see it's on the 90kHz clock which makes my question about rounding kind of pointless. Should this be taking seconds for consistency with other video element APIs? Given baseMediaDecodeTime is going to apply to audio too, using 90k isn't a perfect fit.

imbcmdth: The worker *is* taking seconds. The Transmuxer is taking 90k time. This function is the bridge between the two. I want to keep `videojs-media-sources.js` as transmuxer-agnostic as possible so that it can just worry about seconds where ever possible. Likewise, I'd like the time in the Transmuxer to know about the 90khz clock.

dmlap: I'm not *totally* sold but I can roll with it.

PR #57, opened 2015-11-20
Assign Cue after VTTCue is shimmed

forbesjo:
Cue is being defined before vtt.js is inserted into the page with `emulateTextTracks()`. The consequence of this is that Cue creation will fail for browsers that don't have VTTCue natively.

Unit tests wouldn't catch this issue because this project is using video.js and not video.novtt.js with the emulateTextTracks option.

### DIFF
--- a/src/videojs-media-sources.js
+++ b/src/videojs-media-sources.js
@@ -15,8 +15,6 @@
       createTextTracksIfNecessary,
       addTextTrackData;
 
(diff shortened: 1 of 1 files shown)


dmlap:
The issue was that video.js hadn't necessarily shimmed VTTCue by the time we captured the constructor?

forbesjo:
Yes that is the issue.

PR #59, opened 2015-12-18
Turn videojs-contrib-media-sources into a node.js module with browserify support

BrandonOCasey:
* moved example into its own folder
* removed concat build step (browserify will do this)
* this pull request depends on https://github.com/videojs/mux.js/pull/44 (so we can use mux.js as a node module)

### DIFF
--- a/src/videojs-media-sources.js
+++ b/src/videojs-media-sources.js
@@ -191,9 +191,18 @@ addTextTrackData = function (sourceHandler, captionArray, metadataArray) {
       // MediaSource
       this.sourceBuffers = [];
 
+      // Re-emit MediaSource events on the polyfill
+      [
+        'sourceopen',
+        'sourceclose',
+        'sourceended'
+      ].forEach(function(eventName) {
+        this.mediaSource_.addEventListener(eventName, this.trigger.bind(this));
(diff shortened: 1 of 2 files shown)


dmlap:
@BrandonOCasey given this project is a proper video.js plugin and not just a library like mux.js, I think we should be shooting to get it into line with the new build and code conventions in the [plugin generator](https://github.com/videojs/generator-videojs-plugin).

Gruntfile.js
```diff
@@ -86,89 +96,6 @@ module.exports = function(grunt) {
       }
     }
   });
-
-  grunt.task.registerMultiTask(
```
imbcmdth: The point of these two grunt tasks is to take the transmuxer_worker.js file, inline it's contents (which is admittedly done by browserify now so that is ok) and then create a blob worker and change the `new Worker` line in videojs-media-sources.js so that instead of pointing to a file, it points to an inlined web worker created via a blob object. This allows us to use a web worker without making an extra request to the server for a file (in this case transmuxer_worker.js.)

I'm not sure current build outputs a truly "stand-alone" file with worker without doing at least some of the gymnastics that we previously performed!

PR #61, opened 2015-12-23
Give the closed-caption track a proper label so that it doesn't show up as `undefined`

imbcmdth:


src/transmuxer_worker.js
```diff
@@ -74,6 +74,9 @@ var messageHandlers = {
    * default options if `init` was never explicitly called
    */
   defaultInit: function () {
+    if (transmuxer) {
+      transmuxer.dispose();
```
imbcmdth: This is just a fix that I decided to sneak in here. It wasn't breaking anything but I didn't like the idea of there being the possibility that the *old* transmuxer (if there was one) was still emitting events that we could be pumping through the web worker interface.

src/videojs-media-sources.js
```diff
@@ -62,7 +62,7 @@ createTextTracksIfNecessary = function (sourceBuffer, mediaSource, segment) {
   if (segment.captions &&
       segment.captions.length &&
       !sourceBuffer.inbandTextTrack_) {
-    sourceBuffer.inbandTextTrack_ = mediaSource.player_.addTextTrack('captions');
+    sourceBuffer.inbandTextTrack_ = mediaSource.player_.addTextTrack('captions', 'cc1');
```
imbcmdth: I mean, it makes sense to me!

gkatsev: can we have, or do we support multiple captions?

PR #63, opened 2016-01-11
Duration fixes

imbcmdth:
Only use the locally-stored duration value if the mediaSource has a duration of Infinity.


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "videojs-contrib-media-sources",
-  "version": "2.4.3",
+  "version": "2.4.4",
```
dmlap: Why do these version changes keep sneaking in?

PR #62, opened 2015-12-23
Terminate any sourcebuffer's web workers when closing the media source

imbcmdth:

PR #64, opened 2016-01-15
browserifiy, pluginifiy, and nodeifiy media-sources

BrandonOCasey:
changed everything to es6, split files up, uses new generator conversions
example now follows standards and works
added build step for example, and got example working for real
now following the vjs standards
fixed a lint error with the example
moved the example to its own folder and browserfied it
converted files to use node style, added browserify to grunt build, added browserify to karma unit tests
updated readme to point to new example location
changed everything to es6, split files up, uses new generator conversions
example now follows standards and works
added build step for example, and got example working for real
now following the vjs standards
fixed a lint error with the example
merging recent changes from master
fixing the readme
Terminate any sourcebuffer's web workers when closing the media source
Merge pull request #62 from imbcmdth/terminator
specify the babelify transform so that browserify can require us from other modules
verify that the plugin is registed for a sanity check
fixed function scopes
ignore example modifications
ignore example during linting
fixing lint errors in plugin
ignore safari tests did not work in it before
merged in web worker termination changes, and test


misteroneill:
Holy wall of text, Batman! :laughing:

.npmignore
```diff
@@ -1,10 +1,4 @@
...
-*.log
-tmp
-test
-
-# ignore emacs scratch files
-*~
+!dist/
+!dist-test/
+!docs/
+!es5/
```
misteroneill: You might want to pick up the latest changes in the generator - the `.npmignore` is empty and what gets included in the package is explicitly outlined in `package.json`'s `"files"` field.

TODO.md
```diff
@@ -0,0 +1 @@
+1. Lint tests
```
misteroneill: I feel like this shouldn't be checked in, but rather an issue should be filed.

package.json
```diff
@@ -2,46 +2,103 @@
   "name": "videojs-contrib-media-sources",
   "version": "2.4.3",
   "description": "A Media Source Extensions plugin for video.js",
-  "main": "videojs-media-sources.js",
+  "author": "Brightcove",
```
misteroneill: Very nitpicky, but the generator would set this to `"Brightcove, Inc."`

package.json
```diff
@@ -2,46 +2,103 @@
...
+  "license": "Apache-2.0",
+  "main": "src/plugin.js",
```
misteroneill: This should be `"es5/plugin.js"` because we can't guarantee that people importing this package will be using ES6/Babel.

gkatsev: Not only that but browserify transforms don't run on dependencies and we don't want to force users of the package to use babel themselves.

package.json
```diff
@@ -2,46 +2,103 @@
...
+  "browser": "src/plugin.js",
```
misteroneill: Same here. This should be `"es5/plugin.js"`, I think.

package.json
```diff
@@ -2,46 +2,103 @@
...
+  "keywords": [
+    "videojs",
+    "videojs-plugin"
+  ],
+  "browserify": {
+    "transform": [
+      "browserify-shim",
+      "babelify"
```
misteroneill: The reason this isn't included in here in the generator's output is that the Babel to Browserify process is two steps because we want the `es5/` directory to exist for users who are not using Babel.

TODO.md
BrandonOCasey: going to delete this file, it was my own personal TODO, that is not up to date

package.json
BrandonOCasey: yeah I never changed it back when we talked about it before

src/schedule-tick.js
```diff
@@ -0,0 +1,9 @@
+import FlashMediaSource from './flash-media-source';
+
+const scheduleTick = function(func) {
```
dmlap: I think we have a bit of file-itis going on here :) Couldn't this go directly into flash-media-source.js?

src/transmuxer-worker.js
```diff
@@ -8,35 +8,27 @@
...
-importScripts('../node_modules/mux.js/lib/utils/stream.js');
-importScripts('../node_modules/mux.js/lib/utils/exp-golomb.js');
-importScripts('../node_modules/mux.js/lib/mp4/mp4-generator.js');
-importScripts('../node_modules/mux.js/lib/codecs/aac.js');
-importScripts('../node_modules/mux.js/lib/codecs/h264.js');
-importScripts('../node_modules/mux.js/lib/m2ts/m2ts.js');
-importScripts('../node_modules/mux.js/lib/m2ts/caption-stream.js');
-importScripts('../node_modules/mux.js/lib/m2ts/metadata-stream.js');
-importScripts('../node_modules/mux.js/lib/mp4/transmuxer.js');
+let muxjs = require('mux.js');
```
dmlap: require -> import

src/transmuxer-worker.js
```diff
@@ -8,35 +8,27 @@
...
-  transmuxer.on('data', function (segment) {
-    // transfer ownership of the underlying ArrayBuffer instead of doing a copy to save memory
+const wireTransmuxerEvents = function(transmuxer) {
+  transmuxer.on('data', function(segment) {
+    // transfer ownership of the underlying ArrayBuffer
+    // instead of doing a copy to save memory
     // ArrayBuffers are transferable but generic TypedArrays are not
+    /* eslint-disable max-len */
     // see https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Using_web_workers#Passing_data_by_transferring_ownership_(transferable_objects)
+    /* eslint-enable max-len */
```
dmlap: We could use a shortened URL to get around the line length problem: https://goo.gl/TBDNny

src/transmuxer-worker.js
```diff
@@ -8,35 +8,27 @@
...
-importScripts('../node_modules/mux.js/lib/utils/exp-golomb.js');
-importScripts('../node_modules/mux.js/lib/mp4/mp4-generator.js');
-importScripts('../node_modules/mux.js/lib/codecs/aac.js');
-importScripts('../node_modules/mux.js/lib/codecs/h264.js');
-importScripts('../node_modules/mux.js/lib/m2ts/m2ts.js');
-importScripts('../node_modules/mux.js/lib/m2ts/caption-stream.js');
-importScripts('../node_modules/mux.js/lib/m2ts/metadata-stream.js');
-importScripts('../node_modules/mux.js/lib/mp4/transmuxer.js');
+let muxjs = require('mux.js');
+let gTransmuxer;
```
dmlap: What's the `g` for?

src/transmuxer-worker.js
gkatsev: We theoretically don't need either. Unless it's been changed, vjsstandard should just warn and not fail on long lines.

BrandonOCasey: do we want to be warned about stuff that does not matter every time we lint though? Link shortener would also work

src/transmuxer-worker.js
BrandonOCasey: global, maybe I should just name it globalTransform. wireTransmuxerEvents takes in transmuxer as an argument (what this variable was previously called) which was confusing, and vjsstandard did not like it.

src/schedule-tick.js
BrandonOCasey: file-itis confirmed

gkatsev: Better to have extra files than everything in the same file :)

PR #65, opened 2016-01-16
Fix handling of arraybuffers

imbcmdth:
When casting typed-arrays to raw array-buffers for zero-copy into and out-of the web worker, be sure to keep the original view into the array buffer.

### DIFF
--- a/test/media-sources_test.js
+++ b/test/media-sources_test.js
@@ -44,7 +44,35 @@
       },
       initializeNativeSourceBuffers,
       oldFlashTransmuxer,
-      MockSegmentParser;
+      MockSegmentParser,
+
+      // Create a WebWorker-style message that simulates data being emitted
+      // by the transmuxer
+      createDataMessage = function(type, typedArray, extraObject) {
+        var message = {
+          data: {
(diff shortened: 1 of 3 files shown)


dmlap:
Code looks good but we have one test failure. Hopefully that means we have a test that can easily be modified to ensure we keep passing along `byteOffset` and `byteLength`.

dmlap:
For background, this fix was causing some HLSe streams in contrib-hls. The pkcs7 padding was `subarray()`ed off but the byte window was lost being sent to the transmuxer and the padding was re-added.

test/media-sources_test.js
```diff
@@ -44,7 +44,35 @@
       },
       initializeNativeSourceBuffers,
       oldFlashTransmuxer,
-      MockSegmentParser;
+      MockSegmentParser,
+
+      // Create a WebWorker-style message that simulates data being emitted
+      // by the transmuxer
+      createDataMessage = function(type, typedArray, extraObject) {
```
dmlap: :+1:

dmlap:
LGTM

Issue #58, opened 2015-12-02
Improvement on videojs.MediaSource.supportsNativeMediaSources to handle cases better


muratakbal:
Apparently, implementation of `videojs.MediaSource.supportsNativeMediaSources` only checks if `MediaSource` is available in `window`. But, even if `window.MediaSource` is available, there's a chance that given mimeType is not supported by browser or disabled by user. As an example, Firefox doesn't support mp4 in some operating systems, but still supports media source. In that case, that logic'll choose `HtmlMediaSource` as media source and `nativeMediaSource.addSourceBuffer` will throw `NotSupportedError: Operation is not supported` because mp4 is not supported.

I don't know how hard it is, but somehow if you can check [isTypeSupported](https://developer.mozilla.org/en-US/docs/Web/API/MediaSource/isTypeSupported) in `videojs.MediaSource.supportsNativeMediaSources`, that would be more accurate.

@gkatsev wdyt?

gkatsev:
That sounds right to me. @dmlap @imbcmdth?

dmlap:
Yeah, makes sense to me too.

GeyseR:
Hi, there.
We use Brightcove solution for delivering videos for our users. Recent version of Brightcove player based on videojs library and use this plugin as it's part. And we faced this error before and, unfotunately, after the fix above was merged (i checked that brightcove compiled script on our page has new code from PR) from . Error happens (only in FF, and only for small number of users) when script tries to execute `nativeMediaSource.addSourceBuffer('audio/mp4;codecs="' + this.codecs_[1] + '"')` statement. 
Probably we should check support not only for `'video/mp4;codecs="avc1.4d400d,mp4a.40.2"'` codecs, but for `'video/mp4;codecs="avc1.4d400d"'` and `'audio/mp4;codecs="mp4a.40.2"'`, too (see full fragment below).

Full fragment with adding source buffers:
```
var
        segment = event.data.segment,
        nativeMediaSource = this.mediaSource_.mediaSource_;

      // Cast ArrayBuffer to TypedArray
      segment.data = new Uint8Array(segment.data, event.data.byteOffset, event.data.byteLength);

      // If any sourceBuffers have not been created, do so now
      if (segment.type === 'video') {
        if (!this.videoBuffer_) {
          this.videoBuffer_ = nativeMediaSource.addSourceBuffer('video/mp4;codecs="' + this.codecs_[0] + '"');
        }
[...]

PR #67, opened 2016-02-10
Check MediaSources isTypeSupported with the default codec information…

imbcmdth:
… to allow us to fallback to flash when the platform doesn't natively support H.264 or AAC via MSE.

This happens in Chromium which has no native support for mp4 and in both Firefox and IE which use OS provided (but all too often missing) codecs to decode mp4.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@ language: node_js
 sudo: false
 node_js:
   - "stable"
+addons:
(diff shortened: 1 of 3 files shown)


imbcmdth:
Fixes #58.

forbesjo:
LGTM

misteroneill:
LGTM

PR #70, opened 2016-03-08
Support for two VirtualSourceBuffers (audio and video/combined)

gesinger:
This add support for two VirtualSourceBuffers, one audio only, and one either video only or video and audio combined.

This is a working PR, and is proposed for general comments/critique. It has not been tested thoroughly, as full testing is pending the Video.js audioTracks API.

Note that there is a TODO within the code to switch an event listener from on to one. In running tests, the one listener was hitting the max stack limit, and needs further debugging to discover where the recursion is occurring.

edit: removed "not ready" at beginning

### DIFF
--- a/src/html-media-source.js
+++ b/src/html-media-source.js
@@ -134,7 +134,6 @@ export default class HtmlMediaSource extends videojs.EventTarget {
     let mp4aCodec;
     let avcRegEx = /avc1\.[\da-f]+/i;
     let mp4aRegEx = /mp4a\.\d+.\d+/i;
-    let setActiveSourceBuffers;
 
     // create a virtual source buffer to transmux MPEG-2 transport
     // stream segments into fragmented MP4s
@@ -160,14 +159,11 @@ export default class HtmlMediaSource extends videojs.EventTarget {
       buffer = this.mediaSource_.addSourceBuffer(type);
     }
 
-    setActiveSourceBuffers = () => {
-      this.updateActiveSourceBuffers_();
-      // TODO buffer.one causing call stack exceeded exception
-      buffer.off('updateend', setActiveSourceBuffers);
-    };
-    // for combined audio/video tracks, we can only determine if a source buffer is
-    // active after a completed update (once it has/doesn't have videoTracks)
-    buffer.on('updateend', setActiveSourceBuffers);
+    // For combined audio/video tracks, we can only determine if a source buffer is
+    // active after a completed update (once it has/doesn't have videoTracks).
+    // Once https://github.com/videojs/video.js/issues/2981 is resolved, switch to using
+    // buffer.one instead of buffer.on.
+    buffer.on('updateend', this.updateActiveSourceBuffers_.bind(this));
 
     this.sourceBuffers.push(buffer);
     return buffer;
--- a/src/html-media-source.js
+++ b/src/html-media-source.js
@@ -63,9 +63,17 @@ export default class HtmlMediaSource extends videojs.EventTarget {
       }
     });
 
+    Object.defineProperty(this, 'activeSourceBuffers', {
+      get() {
+        return self.activeSourceBuffers_;
+      }
+    });
+
     // the list of virtual and native SourceBuffers created by this
     // MediaSource
     this.sourceBuffers = [];
+
+    this.activeSourceBuffers_ = [];
 
     // Re-emit MediaSource events on the polyfill
     [
@@ -86,6 +94,7 @@ export default class HtmlMediaSource extends videojs.EventTarget {
       }
 
       self.player_ = videojs(video.parentNode);
+      self.player_.audioTracks().on('change', self.updateActiveSourceBuffers_.bind(self));
     });
 
     // explicitly terminate any WebWorkers that were created
@@ -125,6 +134,7 @@ export default class HtmlMediaSource extends videojs.EventTarget {
     let mp4aCodec;
     let avcRegEx = /avc1\.[\da-f]+/i;
     let mp4aRegEx = /mp4a\.\d+.\d+/i;
+    let setActiveSourceBuffers;
 
     // create a virtual source buffer to transmux MPEG-2 transport
     // stream segments into fragmented MP4s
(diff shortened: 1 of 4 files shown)


src/html-media-source.js
```diff
@@ -145,14 +155,43 @@ export default class HtmlMediaSource extends videojs.EventTarget {
...
+    } else {
+      // delegate to the native implementation
+      buffer = this.mediaSource_.addSourceBuffer(type);
     }
 
-    // delegate to the native implementation
-    buffer = this.mediaSource_.addSourceBuffer(type);
+    setActiveSourceBuffers = () => {
+      this.updateActiveSourceBuffers_();
+      // TODO buffer.one causing call stack exceeded exception
```
dmlap: There's a bug in video.js for this somewhere...

src/html-media-source.js
```diff
@@ -145,14 +155,43 @@ export default class HtmlMediaSource extends videojs.EventTarget {
...
+      buffer.off('updateend', setActiveSourceBuffers);
+    };
+    // for combined audio/video tracks, we can only determine if a source buffer is
+    // active after a completed update (once it has/doesn't have videoTracks)
+    buffer.on('updateend', setActiveSourceBuffers);
```
dmlap: Is there harm in just checking/adjusting the active source buffers on every `updateend`? Cuts down on the state a bit.

src/html-media-source.js
```diff
@@ -145,14 +155,43 @@ export default class HtmlMediaSource extends videojs.EventTarget {
...
+      // We are using an alternate audio track from the default. Since we currently only
+      // support a max of two source buffers, add all of the source buffers (in order).
+      this.sourceBuffers.forEach((sourceBuffer) => {
+        this.activeSourceBuffers_.push(sourceBuffer);
+      });
+    } else {
+      // We are using the combined audio/video stream, so only add the combined source
+      // buffer.
+      this.sourceBuffers.forEach((sourceBuffer) => {
+        if (sourceBuffer.videoTracks && sourceBuffer.videoTracks.length > 0) {
```
dmlap: Could we skip adding `videoTracks` support right now by checking the codec the source buffer was created with?

src/virtual-source-buffer.js
```diff
@@ -154,6 +154,12 @@ export default class VirtualSourceBuffer extends videojs.EventTarget {
         return videojs.createTimeRanges(ranges);
       }
     });
+
+    Object.defineProperty(this, 'videoTracks', {
```
dmlap: Could we expose the `videoTracks` directly instead of going the accessor route?

src/html-media-source.js
gkatsev: https://github.com/videojs/video.js/issues/2981

test/html.test.js
```diff
@@ -665,3 +672,88 @@ QUnit.test('does not wrap mp4 source buffers', function() {
   );
   QUnit.equal(mediaSource.sourceBuffers.length, 2, 'created native buffers');
 });
+
+QUnit.test('can get activeSourceBuffers', () => {
```
dmlap: I'm cool with fat-arrow functions for test definitions but I think we should be consistent within individual files (and we're using `function` defs up to this point).

test/html.test.js
```diff
@@ -665,3 +672,88 @@ QUnit.test('does not wrap mp4 source buffers', function() {
...
+  let mediaSource = new videojs.MediaSource();
+
+  QUnit.ok(Array.isArray(mediaSource.activeSourceBuffers));
```
dmlap: *Technically* active source buffers should be a SourceBufferList so it seems weird to use `Array.isArray` here. What about testing for `length` or something else array-ish?

test/html.test.js
```diff
@@ -665,3 +672,88 @@ QUnit.test('does not wrap mp4 source buffers', function() {
...
+  QUnit.equal(updateCallCount, 2,
+              'active source buffers updated after first updateend of new source buffer');
+
+  sourceBuffer.videoBuffer_.trigger('updateend');
+  QUnit.equal(updateCallCount, 2,
+              'active source buffers not updated after second updateend of new source ' +
+              'buffer');
+});
+
+// running under the assumption that we will have a max of two source buffers
```
dmlap: I don't see any assumptions in this test that strike me as untrue if we allowed more than two source buffers (given the test only creates two). Am I missing something?

src/html-media-source.js
gesinger: Should be fine to run every time. I'll put a note above though to convert to "one" in the future (saves a little bit of processing).

src/html-media-source.js
gesinger: We are just using videoTracks to see what source buffer represents video/combined, since we don't have a way of identifying it otherwise (I may be reading it wrong, but it seems that all source buffers are added with the mimeType of just 'video/mp2t' with no codecs specified, from https://github.com/videojs/videojs-contrib-hls/blob/my-rebase/src/segment-loader.js#L145 ).

src/virtual-source-buffer.js
gesinger: We can do that, but we'd have to point a new local property called videoTracks at videoBuffer_.videoTrack.

test/html.test.js
gesinger: Good point. Will change. And you're doubly right because tests are better suited to not using fat arrows, as we are using "this" a lot in them.

test/html.test.js
gesinger: It's kinda funny, but that was exactly why I used Array.isArray in this test. I probably should've made a comment or added it to the test title, but since we are going with the simplified route of using an array instead of a SourceBufferList, I wanted to put it in the tests as a kind of "documentation" to show that we meant to do that and have assumptions based on it.

test/html.test.js
gesinger: You're right. Will remove.

src/html-media-source.js
dmlap: That's a bug in segment loader, IMO

gesinger: I agree, and I think the codecs used to be passed. Should I add a TODO to use the codecs? Or can we not guarantee that users will pass them?

dmlap: Let's go with the TODO-- I think we should require them.

src/virtual-source-buffer.js
dmlap: That doesn't seem bad to me... am I missing a downside?

dmlap:
LGTM

PR #71, opened 2016-03-11
lock webworkify to 1.0.2 to fix ie8 '.default' usage error

BrandonOCasey:
webworkify uses exp.default and the default part throws an error in IE8
added webworkify to dependancies rather than dev dependancies


imbcmdth:
LGTM

PR #72, opened 2016-03-13
Remove global videojs, prefer importing module.

hallelujah:
window.videojs should not be used


src/videojs-contrib-media-sources.js
```diff
@@ -1,7 +1,7 @@
 import FlashMediaSource from './flash-media-source';
 import HtmlMediaSource from './html-media-source';
-// import videojs from 'video.js';
-let videojs = window.videojs;
+import videojs from 'video.js';
+// let videojs = window.videojs;
```
forbesjo: Could you remove this line?

hallelujah:
@forbesjo  Hmm it seems vjsstandard is failing in another place in the code

PR #74, opened 2016-03-23
better way to determine which sourceBuffer to enable

BrandonOCasey:
fix a bug where we would not remove audioBuffer if audio was disabled
better way to determine audio and video codec

### DIFF
--- a/src/html-media-source.js
+++ b/src/html-media-source.js
@@ -1,23 +1,43 @@
 import videojs from 'video.js';
 import VirtualSourceBuffer from './virtual-source-buffer';
+import {isAudioCodec, isVideoCodec} from './codec-check';
 
 // Replace the old apple-style `avc1.<dd>.<dd>` codec string with the standard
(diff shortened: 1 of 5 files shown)


src/html-media-source.js
```diff
@@ -172,51 +172,46 @@ export default class HtmlMediaSource extends videojs.EventTarget {
...
-        /* eslinst-disable */
-        // TODO once codecs are required, we can switch to using the codecs to determine
-        //      what stream is the video stream, rather than relying on videoTracks
-        /* eslinst-enable */
-        if (sourceBuffer.videoCodec_) {
-          this.activeSourceBuffers_.push(sourceBuffer);
-        }
-      });
-    }
+      this.activeSourceBuffers_.push(sourceBuffer);
```
gesinger: This looks like it will add all of the virtual source buffers to activeSourceBuffers, even if only one virtual source buffer (the combined one) is enabled.

PR #75, opened 2016-03-25
Fix Unit tests

BrandonOCasey:
fix a bug where we would not remove audioBuffer if audio was disabled
better way to determine audio and video codec
only add audioOnly buffer to active source buffers when its enabled
ran the vjsstandard
got all tests working
improved realsourcebuffer creation logic
added two tests for video only and audio only
consolidated codec regex
improved content type parsing
added unit tests for video/audio only codecs

### DIFF
--- a/src/html-media-source.js
+++ b/src/html-media-source.js
@@ -1,23 +1,23 @@
 import videojs from 'video.js';
 import VirtualSourceBuffer from './virtual-source-buffer';
+import {isAudioCodec, isVideoCodec, parseContentType} from './codec-utils';
 
 // Replace the old apple-style `avc1.<dd>.<dd>` codec string with the standard
 // `avc1.<hhhhhh>`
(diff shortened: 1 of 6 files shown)


src/html-media-source.js
```diff
@@ -165,51 +179,57 @@ export default class HtmlMediaSource extends videojs.EventTarget {
...
+    // Retain the reference but empty the array
+    this.activeSourceBuffers_.length = 0;
 
+    let combined = 'enable';
+    let audioOnly = 'disable';
+
+    // TODO: find a better way to determine which sourcebuffers audio
+    // needs to be enabled this method relies on the track with kind 'main'
+    // being in the combined sourcebuffer. It is possible for the main track
+    // to have no audio and have a seprate track be the main audio
```
dmlap: I attempted to write up how to do this in a gist: https://gist.github.com/dmlap/26809ba67487ad413d47

PR #80, opened 2016-04-04
fix currently broken unit tests

BrandonOCasey:
added browserify debug to karma

### DIFF
--- a/src/html-media-source.js
+++ b/src/html-media-source.js
@@ -17,7 +17,7 @@ const translateLegacyCodecs = function(codecs) {
 
 export default class HtmlMediaSource extends videojs.EventTarget {
   constructor() {
-    super(videojs.EventTarget);
+    super();
     let property;
 
     this.mediaSource_ = new window.MediaSource();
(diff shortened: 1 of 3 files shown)


src/html-media-source.js
```diff
@@ -211,7 +216,7 @@ export default class HtmlMediaSource extends videojs.EventTarget {
     // Automatically disable the audio on the first source buffer if
     // a second source buffer is ever created
     if (this.sourceBuffers.length !== 0) {
-      this.sourceBuffers[0].disableAudio();
+      this.sourceBuffers[0].audioDisabled_ = true;
```
imbcmdth: Why was this changed?

src/html-media-source.js
```diff
@@ -133,6 +133,8 @@ export default class HtmlMediaSource extends videojs.EventTarget {
     // capture the associated player when the MediaSource is
     // successfully attached
     this.on('sourceopen', (event) => {
+      // TODO: find a better way to get the player this MediaSoruce
+      // is attached to
```
imbcmdth: There isn't a better way unless they change MSE :smile: you might as well remove the comment.

src/html-media-source.js
BrandonOCasey: for the test in html called "does not wrap mp4 source buffers" it was throwing an error, since native source buffers will not have a `disableAudio()` function, so I changed it to just change to property which can be created but not used on native source buffers. We may even want to change all `disableAudio` and `enableAudio` calls in html-media-source to statements like this?

imbcmdth:
Pulling this in via #81

PR #82, opened 2016-04-05
save information about the current audio stream on mediaSource

BrandonOCasey:
take that information and emit an audioinfochanged if we detect that it has changed

### DIFF
--- a/src/html-media-source.js
+++ b/src/html-media-source.js
@@ -264,48 +264,4 @@ export default class HtmlMediaSource extends videojs.EventTarget {
     this.sourceBuffers.push(buffer);
     return buffer;
   }
-
-  /**
-   * add a property to the VSB indicating the current audio/video/text stream
-   * properties (sample rate, channels, etc) and fire an event
-   * when the info changes
-   *
-   * @private
-   * @param {String} type the type of stream to change info on (audio, video, text)
-   * @param {Object} info info about the currently muxed audio/video/text stream
-   */
-  changeInfo_(type, info) {
-    let propName = `${type}Info_`;
-
-    if (!this[propName]) {
-      this[propName] = info;
-      return;
-    }
-
-    let change = () => {
-      // copy current properties to another object
-      let old = videojs.mergeOptions({}, this[propName]);
-
-      this[propName] = info;
-      this.trigger({
-        revert: () => this[propName] = old,
-        type: `${type}infochanged`
-      });
-    };
-
-    // if we have a different number of elements
-    // something has changed
-    if (Object.keys(this[propName]).length !== Object.keys(info).length) {
-      change();
-      return;
-    }
-
-    for (let prop in info) {
-      if (this[propName][prop] !== info[prop]) {
-        change();
-        return;
-      }
-    }
-  }
-
 }
--- a/src/html-media-source.js
+++ b/src/html-media-source.js
@@ -264,4 +264,41 @@ export default class HtmlMediaSource extends videojs.EventTarget {
     this.sourceBuffers.push(buffer);
     return buffer;
   }
+
+  /**
+   * add a property to the VSB indicating the current audio/video/text stream
+   * properties (sample rate, channels, etc) and fire an event
+   * when the info changes
+   *
+   * @private
+   * @param {String} type the type of stream to change info on (audio, video, text)
+   * @param {Object} info info about the currently muxed audio/video/text stream
+   */
+  changeInfo_(type, info) {
+    let propName = `${type}Info_`;
+    let old = {};
+
+    if (!this[propName]) {
+      this[propName] = info;
+      return;
+    }
+
+    // copy current properties to another object
+    for (let prop in this[propName]) {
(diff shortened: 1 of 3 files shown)


BrandonOCasey:
This one still needs unit tests

src/virtual-source-buffer.js
```diff
@@ -348,6 +352,10 @@ export default class VirtualSourceBuffer extends videojs.EventTarget {
       return segmentObj;
     }, sortedSegments);
 
+    if (sortedSegments.audio.info) {
+      this.mediaSource_.changeInfo_('audio', sortedSegments.audio.info);
+    }
```
imbcmdth: Aren't we still going to try the append even though it will likely fail with a decode error?

BrandonOCasey: We abort in hls by switching off of this sourebuffer but only in Firefox (so it never gets appended to an active sourebuffer) at least that is what I saw

src/html-media-source.js
```diff
@@ -264,4 +264,41 @@ export default class HtmlMediaSource extends videojs.EventTarget {
...
+
+    // copy current properties to another object
+    for (let prop in this[propName]) {
+      old[prop] = this[propName][prop];
+    }
+
+    for (let prop in info) {
+      if (this[propName][prop] !== info[prop]) {
+        this[propName] = info;
+        this.trigger({
```
gesinger: Are we using the old and new data anywhere? If not, it might make for cleaner code to avoid copying old and passing old and new.

src/html-media-source.js
```diff
@@ -264,4 +264,41 @@ export default class HtmlMediaSource extends videojs.EventTarget {
...
+   *
+   * @private
+   * @param {String} type the type of stream to change info on (audio, video, text)
+   * @param {Object} info info about the currently muxed audio/video/text stream
+   */
+  changeInfo_(type, info) {
+    let propName = `${type}Info_`;
+    let old = {};
+
+    if (!this[propName]) {
```
gesinger: Should we also check that info and this[propName] have the same number of elements? If info is empty, it looks like everything would be fine (unless we expect that behavior).

src/html-media-source.js
BrandonOCasey: we use old in HLS to reset the mediasource audio info to the previous value on firefox, so that we don't re-trigger changeInfo again when we switch back to the main audio.

gesinger: I see what you mean. It still seems like it is a bit hackey (since HLS ends up knowing the internals and private members for HtmlMediaSource to make the change: `this.masterPlaylistController_.mediaSource.audioInfo_ = e.old`).

One option, though it is still suboptimal, would be to pass a callback, so that HLS can just call "e.revertAudioInfo()."

BrandonOCasey: the callback sounds like a better solution

gesinger:
LGTM

src/html-media-source.js
```diff
@@ -264,4 +264,48 @@ export default class HtmlMediaSource extends videojs.EventTarget {
     this.sourceBuffers.push(buffer);
     return buffer;
   }
+
+  /**
+   * add a property to the VSB indicating the current audio/video/text stream
```
dmlap: I'd avoid acronyms (i.e. VSB) in our documentation or define them before using if it's really onerous.

test/html.test.js
```diff
@@ -930,3 +930,133 @@ function() {
     'active source buffers ends with audio source buffer');
 
 });
+
+QUnit.test('calls changeInfo_ on mediaSource when info is given for audio', function() {
```
dmlap: Seems like this is covered by the test below. Any background on why the extra check is needed?

src/html-media-source.js
```diff
@@ -264,4 +264,48 @@ export default class HtmlMediaSource extends videojs.EventTarget {
...
+      return;
+    }
+
+    let change = () => {
+      // copy current properties to another object
+      let old = videojs.mergeOptions({}, this[propName]);
+
+      this[propName] = info;
+      this.trigger({
+        revert: () => this[propName] = old,
```
dmlap: What about using [preventDefault()](https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault) to trigger this action?

dmlap: Also, if you stored the audio info in this event, instead of on the MediaSource itself, you wouldn't need to revert anything.

test/html.test.js
BrandonOCasey: One is verifying that `changeInfo_` is always called when there is info on an audio segment. The other verifies that an event is triggered when the info of a track does actually change.

src/html-media-source.js
BrandonOCasey: hmmm it might make more sense to do it like that. We trigger an audioinfo event and have HLS determine wether the info for the stream has changed, probably by doing something similar to this function, on every event.

dmlap:
LGTM

PR #83, opened 2016-04-06
Add isTypeSupported method existing checker

defli:
if MediaSource object exists but isTypeSupported method doesn't exist, script and videoJS will be broken.

This issue can be reproduced in Opera TV OS or Philips Android TV which uses Opera as built-in HTML5 application browser.

### DIFF
--- a/src/videojs-contrib-media-sources.js
+++ b/src/videojs-contrib-media-sources.js
@@ -36,7 +36,7 @@ const open = function(msObjectURL, swfId) {
 // Check to see if the native MediaSource object exists and supports
 // an MP4 container with both H.264 video and AAC-LC audio
 const supportsNativeMediaSources = function() {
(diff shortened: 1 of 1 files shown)


gkatsev:
And those browsers have MediaSources but not `isTypeSupported` method?

defli:
@gkatsev Yes, There is the ss http://imgur.com/wxVZbbO You can test with Opera TV emulator v3.4 (standalone) http://www.operasoftware.com/products/tv/tv-emulator

defli:
@gkatsev you need to http://www.opera.com/dragonfly/ tool but dragonfly only included to opera 12 or below versions. Because Opera has started to use chromium after 12.XX version.

defli:
@gkatsev did you take a look?

forbesjo:
This fix looks fine, merging.

PR #84, opened 2016-04-08
Fix updateactive

imbcmdth:
Repair the completely broken logic that resulted from replacing enableAudio and disableAudio functions on VirtualSourceBuffers with a boolean property. The tests were passing completely by accident even though the logic was inverted in two cases.

Also includes a tiny fix that forces `VirtualSourceBuffer#updating` to always return a boolean instead of sometimes returning `undefined`.

### DIFF
--- a/src/html-media-source.js
+++ b/src/html-media-source.js
@@ -79,21 +79,21 @@ export default class HtmlMediaSource extends videojs.EventTarget {
       // Retain the reference but empty the array
       this.activeSourceBuffers_.length = 0;
 
-      // By default, disable the audio in the combined virtual source buffer
-      // and enable the audio-only source buffer.
-      let combined = true;
-      let audioOnly = false;
+      // By default, the audio in the combined virtual source buffer is enabled
+      // and the audio-only source buffer (if it exists) is disabled.
+      let combined = false;
+      let audioOnly = true;
 
       // TODO: maybe we can store the sourcebuffers on the track objects?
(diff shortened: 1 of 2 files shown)


src/html-media-source.js
```diff
@@ -79,6 +79,8 @@ export default class HtmlMediaSource extends videojs.EventTarget {
       // Retain the reference but empty the array
       this.activeSourceBuffers_.length = 0;
 
+      // By default, disable the audio in the combined virtual source buffer
+      // and enable the audio-only source buffer.
       let combined = true;
```
BrandonOCasey: `combined` should start out `false` and `audioOnly` should start out `true`, since they represent `audioDisabled_` and that is the default.

src/html-media-source.js
```diff
@@ -87,7 +89,9 @@ export default class HtmlMediaSource extends videojs.EventTarget {
       for (let i = 0; i < this.player_.audioTracks().length; i++) {
         let track = this.player_.audioTracks()[i];
 
-        if (track.enabled && track.kind !== 'main') {
+        if (track.enabled && track.kind === 'main') {
+          // The enabled track is a combined audio and video track so we'll enable the
+          // combined source buffer and disable the audio-only source buffer.
           combined = false;
```
BrandonOCasey: these need to be switched as well, maybe we should change audioDisabled_ to audioEnabled_ to make this whole thing easier?

PR #85, opened 2016-04-26
Adding support for video and audio only playlists

imbcmdth:
Defer the creation of the underlying real SourceBuffers until one of two conditions is fulfilled:
1. We get data back from the transmuxer so that we know the track layout of the source.
2. a second VirtualSourceBuffer is created meaning that both video and audio source buffers are necessary.

### DIFF
--- a/src/html-media-source.js
+++ b/src/html-media-source.js
@@ -37,12 +37,12 @@ export default class HtmlMediaSource extends videojs.EventTarget {
     super();
     let property;
 
-    this.mediaSource_ = new window.MediaSource();
+    this.nativeMediaSource_ = new window.MediaSource();
     // delegate to the native MediaSource's methods by default
-    for (property in this.mediaSource_) {
+    for (property in this.nativeMediaSource_) {
       if (!(property in HtmlMediaSource.prototype) &&
-          typeof this.mediaSource_[property] === 'function') {
-        this[property] = this.mediaSource_[property].bind(this.mediaSource_);
+          typeof this.nativeMediaSource_[property] === 'function') {
+        this[property] = this.nativeMediaSource_[property].bind(this.nativeMediaSource_);
       }
     }
 
@@ -55,12 +55,12 @@ export default class HtmlMediaSource extends videojs.EventTarget {
         if (this.duration_ === Infinity) {
           return this.duration_;
         }
-        return this.mediaSource_.duration;
+        return this.nativeMediaSource_.duration;
       },
       set(duration) {
         this.duration_ = duration;
         if (duration !== Infinity) {
-          this.mediaSource_.duration = duration;
+          this.nativeMediaSource_.duration = duration;
           return;
         }
       }
@@ -68,15 +68,15 @@ export default class HtmlMediaSource extends videojs.EventTarget {
     Object.defineProperty(this, 'seekable', {
       get() {
         if (this.duration_ === Infinity) {
-          return videojs.createTimeRanges([[0, this.mediaSource_.duration]]);
+          return videojs.createTimeRanges([[0, this.nativeMediaSource_.duration]]);
         }
-        return this.mediaSource_.seekable;
+        return this.nativeMediaSource_.seekable;
       }
     });
 
     Object.defineProperty(this, 'readyState', {
       get() {
-        return this.mediaSource_.readyState;
+        return this.nativeMediaSource_.readyState;
       }
     });
 
@@ -156,7 +156,7 @@ export default class HtmlMediaSource extends videojs.EventTarget {
       'sourceclose',
       'sourceended'
     ].forEach(function(eventName) {
-      this.mediaSource_.addEventListener(eventName, this.trigger.bind(this));
+      this.nativeMediaSource_.addEventListener(eventName, this.trigger.bind(this));
     }, this);
 
     // capture the associated player when the MediaSource is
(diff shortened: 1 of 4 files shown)


src/virtual-source-buffer.js
```diff
@@ -247,29 +241,33 @@ export default class VirtualSourceBuffer extends videojs.EventTarget {
...
-      // a videoCodec. we will add seprate event handlers
-      // for audio only after this block of code
-      if (!this.videoCodec_) {
-        return;
+      // If the mediasource already has a SourceBuffer for the codec
+      // use that
+      if (this.mediaSource_[`${type}Buffer_`]) {
+        buffer = this.mediaSource_[`${type}Buffer_`];
+      } else {
+        buffer = this.nativeMediaSource.addSourceBuffer(
```
dmlap: What in the world do we have two names (`this.mediaSource_` and `this.nativeMediaSource`) for the same underlying MediaSource? That is crazy.

imbcmdth: `this.mediaSource_` is our HtmlMediaSource and `this.nativeMediaSource` is the native, real one. Having two separate properties is much easier to understand that doing `this.mediaSource_.mediaSource_` to get to the native implementation.

I'm going to add an underscore to nativeMediaSource since it shouldn't be used outside of this module.

src/virtual-source-buffer.js
```diff
@@ -247,29 +240,33 @@ export default class VirtualSourceBuffer extends videojs.EventTarget {
...
       if (!this[`${type}Codec_`]) {
         return;
       }
 
-      // if the mediasource already has a buffer for the codec
-      // use that
-      if (this.mediaSource_[`${type}Buffer_`]) {
-        this[`${type}Buffer_`] = this.mediaSource_[`${type}Buffer_`];
+      // Do nothing if a SourceBuffer of this type already exists
+      if (this[`${type}Buffer_`]) {
```
BrandonOCasey: Merge this if statement with the above one, as they do the same thing

src/virtual-source-buffer.js
```diff
@@ -410,6 +396,19 @@ export default class VirtualSourceBuffer extends videojs.EventTarget {
       return segmentObj;
     }, sortedSegments);
 
+    // Create the real source buffers if they don't exist by now
+    if (!this.videoBuffer_ && !this.audioBuffer_) {
+      if (sortedSegments.video.bytes === 0) {
+        this.videoCodec_ = null;
```
BrandonOCasey: maybe add a comment stating something like "If we don't have any video data here then we want to ignore the codec, if we have one, as its wrong"

test/html.test.js
```diff
@@ -980,3 +1021,39 @@ QUnit.test('audio segments with info trigger audioinfo event', function() {
   QUnit.equal(infoEvents.length, 2, 'audio info should trigger');
   QUnit.deepEqual(infoEvents[1].info, newinfo, 'audio info = muxed info');
 });
+
+QUnit.test('creates native SourceBuffers immediately if a second ' +
+  'VirtualSourceBuffer is created', function() {
```
BrandonOCasey: this line needs to be spaced over so that it matches the indent of the line above it: 
``` JavaScript
QUnit.test('fooo bar' +
                 'bar foo', function() {
```

src/virtual-source-buffer.js
imbcmdth: I feel that the template strings, while clever and a nice way to reduce the code, make this code hard enough to reason about that I wanted to keep the conditions under which we exit early explicit. Also there is no good way to have two comments if we combine the conditions into one expression.

src/virtual-source-buffer.js
imbcmdth: Good idea!

test/html.test.js
imbcmdth: Ahhhh!

src/virtual-source-buffer.js
dmlap: I think we haven't properly setup our object-ownership hierarchies if we find ourselves regularly interactive with sub-sub-properties but am ok with coming back to this later. Thank you for the rename.

src/html-media-source.js
```diff
@@ -231,10 +231,22 @@ export default class HtmlMediaSource extends videojs.EventTarget {
...
+    // Create a VirtualSourceBuffer to transmux MPEG-2 transport
     // stream segments into fragmented MP4s
     if (parsedType.type === 'video/mp2t') {
-      // default codecs
+      if (this.sourceBuffers.length !== 0) {
+        // If another VirtualSourceBuffer already exists, then we are creating a
+        // SourceBuffer for an alternate audio track and therefore we know that
+        // the source has both an audio and video track.
+        // That means we should trigger the manual creation of the real
+        // SourceBuffers instead of waiting for the transmuxer to return data
```
dmlap: Is this comment inaccurate or are we forbidding creating an audio VSB *first* and then creating the video VSB?

imbcmdth: The built in assumption is that the first VSB is muxed unless we are told otherwise. If we create a second source buffer, it is fairly safe to assumed that it is audio-only. This assumption is built into HLS and MS right now.

Internally, we have a way transfer (or share) ownership of native SourceBuffers that are for audio data but we do not have a way to transfer (or share) ownership of native video SourceBuffers.

PR #86, opened 2016-04-28
Fixed a bug where we weren't forcing the creation of real sourcebuffers on the second virtual sourcebuffer

imbcmdth:
Fixing a bug with #85 where we were creating the real SourceBuffers for the first VirtualSourceBuffer but still deferring the creation or assignment of the second VirtualSourceBuffer's real source buffers until after it received data back from the transmuxer.

This was causing a problem when switching between multiple audio tracks because we weren't able to clear the real source buffer to allow us to switch tracks immediately.


gesinger:
LGTM

PR #89, opened 2016-05-23
Guard against older versions of Video.js that may not not have an `audioTracks` property on the player

imbcmdth:
…


dmlap:
Waiting for Travis but LGTM

PR #90, opened 2016-05-25
Be more selective about mux.js imports

dmlap:
Import the transmuxers more selectively. This slims down the dist from 89K to 75K on my machine, about 20%.

### DIFF
--- a/src/flash-source-buffer.js
+++ b/src/flash-source-buffer.js
@@ -2,7 +2,7 @@
  * @file flash-source-buffer.js
  */
 import videojs from 'video.js';
-import muxjs from 'mux.js';
+import flv from 'mux.js/lib/flv';
 import removeCuesFromTrack from './remove-cues-from-track';
 import createTextTracksIfNecessary from './create-text-tracks-if-necessary';
 import addTextTrackData from './add-text-track-data';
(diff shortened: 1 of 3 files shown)


dmlap:
Obligatory screenshot:
screen shot 2016-05-25 at 6 52 58 pm

Not an earth-shattering savings but it's not complicated either.

src/flash-source-buffer.js
```diff
@@ -412,4 +412,3 @@ export default class FlashSourceBuffer extends videojs.EventTarget {
     return tags;
   }
 }
-
```
imbcmdth: Did you remove the trailing newline? It's hard to tell if that is what was happening here.

imbcmdth:
Surprised how easy this was! Technically it doesn't even require a new version of mux.js, right? 

Either way, LGTM.

src/flash-source-buffer.js
gkatsev: It seems like it was an extra empty line.

src/flash-source-buffer.js
```diff
@@ -2,7 +2,7 @@
  * @file flash-source-buffer.js
  */
 import videojs from 'video.js';
-import muxjs from 'mux.js';
+import flv from 'mux.js/lib/flv';
```
gkatsev: What about making `mux.js` have a top-level `flv` file that we can use instead. So, that we do `import flv from 'mux.js/flv'`. I think it'll be a bit cleaner and signal that it more explicitly inside of muxjs that we shouldn't be touching that file willy nilly. As is, importing directly into `lib` feels dirty for me.

dmlap:
Everyone agrees with @gkatsev that it would be nice to get the `lib/` out of the import but we couldn't agree on a good way to do it. Leaving it as-is for now and hopefully someone will come up with a better solution later.

dmlap:
@imbcmdth technically no, this didn't require a new release of mux.js. I did split out the inspectors though, which we don't use at runtime in this project.

PR #92, opened 2016-06-06
Create and prepend init segments from serialized track information

gesinger:
Made the test files smaller by removing the samples. I may add some back in to test deleting the property.


test/data/audio-tracks.json
```diff
@@ -0,0 +1,18 @@
+[
+  {
```
BrandonOCasey: I don't know if it matters, but this JSON is invalid due to not double quoting all keys and non-number/boolean values

gesinger: Yeah, it seems to work within our tests, but I'm happy to switch to double quoting if we want more compliant JSON.

PR #93, opened 2016-06-22
Flash fires loadedmetada

tenacex:


src/flash-source-buffer.js
```diff
@@ -169,6 +169,12 @@ export default class FlashSourceBuffer extends videojs.EventTarget {
     };
 
     chunkInData();
+
+    // fire loaded metadata on first segment append
+    if (!this.firstAppendHappened) {
```
gkatsev: is it possible to do this without another state property?

tenacex: The only way I could think of doing it was firing it when we append the encoded flv header. Unfortunately it was throwing off some tests. We don't have anything keeping track of the number of segments we have added to the buffer, and I can't check if the sourceBuffer is empty because of the already appended flv header. Definitely open to suggestions on a better way to do it, but currently haven't had much luck.

gesinger: From https://www.w3.org/TR/html51/semantics-embedded-content.html#eventdef-media-loadedmetadata , do we have enough information to fire loadedmetadata at appendBuffer, or do we have to wait until the buffer is finished appending?

tenacex: I believe it has already finished appending when I call it. Above we call `chunkInData`, which I believe is blocking until we have push the bytes into the source buffer.

gesinger: Isn't scheduleTick within chunkInData using setTimeout to make it asynchronous?

imbcmdth: Firing this after we have appended **all** of the data makes the most sense. It is better to fire these kind of events later rather than too early.

PR #98, opened 2016-08-15
Increase the values of the flash constants to push more data per chunk-interval

imbcmdth:
* Increase chunk-push time to 10ms from 4ms
* Increase minimum size of a chunk to 4kb
* Increase the max size of a chunk to 4mb


dmlap:
LGTM

PR #100, opened 2016-08-23
updating the metadata cues

ruchigupta19:
Normalized the ID3 behavior to follow Safari's implementation.
On each segment:
1) Create cues from the id3 tags that start at the proper time and end at the start of the next id3 in time
2) The last cue should have an end time that is very large (MAX_NUMBER)
3) When more id3 tags are loaded, set the end time of any cues that overlap the first id3 tag to end at the start time of the first id3 tags

### DIFF
--- a/src/add-text-track-data.js
+++ b/src/add-text-track-data.js
@@ -63,22 +63,42 @@ const addTextTrackData = function(sourceHandler, captionArray, metadataArray) {
     }, sourceHandler);
   }
 
+
   if (metadataArray) {
+
+
     metadataArray.forEach(function(metadata) {
       let time = metadata.cueTime + this.timestampOffset;
+      let videoDuration = Number.isFinite(this.mediaSource_.duration) ?
+        this.mediaSource_.duration :
+        Number.MAX_VALUE;
 
       metadata.frames.forEach(function(frame) {
         let cue = new Cue(
-            time,
-            time,
-            frame.value || frame.url || frame.data || '');
+          time,
+          videoDuration,
+          frame.value || frame.url || frame.data || '');
 
         cue.frame = frame;
         cue.value = frame;
         deprecateOldCue(cue);
+
         this.metadataTrack_.addCue(cue);
       }, this);
     }, sourceHandler);
+
+    /** Updating the metadeta cues so that
+     * the endTime of each cue is the startTime of the next cue
+     * the endTime of last cue is the duration of the video
+     */
+    if (sourceHandler.metadataTrack_ && sourceHandler.metadataTrack_.cues) {
+      for (let i = 0; i < sourceHandler.metadataTrack_.cues.length - 1; i++) {
+
+        if (sourceHandler.metadataTrack_.cues[i].endTime != sourceHandler.metadataTrack_.cues[i + 1].startTime) {
+          sourceHandler.metadataTrack_.cues[i].endTime = sourceHandler.metadataTrack_.cues[i + 1].startTime;
+        }
+      }
+    }
   }
 };
 
--- a/src/add-text-track-data.js
+++ b/src/add-text-track-data.js
@@ -96,20 +96,19 @@ const addTextTrackData = function(sourceHandler, captionArray, metadataArray) {
      * the endTime of last cue is the duration of the video
      */
     if (sourceHandler.metadataTrack_ && sourceHandler.metadataTrack_.cues) {
-      for (let i = 0; i <= sourceHandler.metadataTrack_.cues.length - 1; i++) {
-        if (i === sourceHandler.metadataTrack_.cues.length - 1) {
-          sourceHandler.metadataTrack_.cues[i].endTime = videoDuration;
+      let cues = sourceHandler.metadataTrack_.cues;
+
+      for (let i = 0; i <= cues.length - 1; i++) {
+        if (i === cues.length - 1) {
+          cues[i].endTime = videoDuration;
         } else {
-          sourceHandler.metadataTrack_.cues[i].endTime =
-          sourceHandler.metadataTrack_.cues[i + 1].startTime;
+          cues[i].endTime = cues[i + 1].startTime;
         }
       }
+      sourceHandler.mediaSource_.on('sourceended', (event) => {
+        cues[cues.length - 1].endTime = videoDuration;
+      });
     }
-    sourceHandler.mediaSource_.on('sourceended', (event) => {
-      let numberOfCues = sourceHandler.metadataTrack_.cues.length;
-
-      sourceHandler.metadataTrack_.cues[numberOfCues - 1].endTime = videoDuration;
-    });
   }
 };
 
--- a/src/add-text-track-data.js
+++ b/src/add-text-track-data.js
@@ -116,4 +116,7 @@ const addTextTrackData = function(sourceHandler, captionArray, metadataArray) {
   }
 };
 
-export default addTextTrackData;
+export default {
+  addTextTrackData,
+  durationOfVideo
+};
--- a/src/add-text-track-data.js
+++ b/src/add-text-track-data.js
@@ -98,13 +98,11 @@ const addTextTrackData = function(sourceHandler, captionArray, metadataArray) {
     if (sourceHandler.metadataTrack_ && sourceHandler.metadataTrack_.cues) {
       let cues = sourceHandler.metadataTrack_.cues;
 
-      for (let i = 0; i <= cues.length - 1; i++) {
-        if (i === cues.length - 1) {
-          cues[i].endTime = videoDuration;
-        } else {
-          cues[i].endTime = cues[i + 1].startTime;
-        }
+      for (let i = 0; i < cues.length - 1; i++) {
+        cues[i].endTime = cues[i + 1].startTime;
       }
+      cues[cues.length - 1].endTime = videoDuration;
+
       sourceHandler.mediaSource_.on('sourceended', (event) => {
         cues[cues.length - 1].endTime = videoDuration;
       });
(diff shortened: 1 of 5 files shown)


src/add-text-track-data.js
```diff
@@ -66,19 +66,38 @@ const addTextTrackData = function(sourceHandler, captionArray, metadataArray) {
...
+        this.mediaSource_.duration :
+        Number.MAX_VALUE;
 
       metadata.frames.forEach(function(frame) {
         let cue = new Cue(
-            time,
-            time,
-            frame.value || frame.url || frame.data || '');
+          time,
+          videoDuration,
```
imbcmdth: This will generate a `cuechange` event for each cue that is added before the `currentTime`. It might make more sense to start with zero-length cues and extend their durations instead of starting at full-length and shortening them below.

src/add-text-track-data.js
```diff
@@ -66,19 +66,38 @@ const addTextTrackData = function(sourceHandler, captionArray, metadataArray) {
...
         this.metadataTrack_.addCue(cue);
       }, this);
     }, sourceHandler);
+
+    /** Updating the metadeta cues so that
+     * the endTime of each cue is the startTime of the next cue
+     * the endTime of last cue is the duration of the video
+     */
+    if (sourceHandler.metadataTrack_ && sourceHandler.metadataTrack_.cues) {
+      for (let i = 0; i < sourceHandler.metadataTrack_.cues.length - 1; i++) {
```
imbcmdth: My only reservation about this is that we will be iterating over all the cues every segment append. There may not ever be a large number of cues but this algorithm has a complexity of O(n^2) which has scary implications in extreme cases.

TBH, I'm not sure there is a good way to make this take less time without vastly complicating the algorithm but I felt like pointing this out since that kind of complexity can cause performance issues if you aren't careful.

src/add-text-track-data.js
```diff
@@ -66,19 +66,38 @@ const addTextTrackData = function(sourceHandler, captionArray, metadataArray) {
...
+
+        if (sourceHandler.metadataTrack_.cues[i].endTime !==
+          sourceHandler.metadataTrack_.cues[i + 1].startTime) {
+          sourceHandler.metadataTrack_.cues[i].endTime =
+            sourceHandler.metadataTrack_.cues[i + 1].startTime;
+        }
+      }
+    }
```
imbcmdth: I think we will need to add code to `endOfStream` that will set the endpoint of the final cue in the `cues` array to have an `endTime` that is equal to the video's duration.

src/add-text-track-data.js
```diff
@@ -66,19 +66,38 @@ const addTextTrackData = function(sourceHandler, captionArray, metadataArray) {
   if (metadataArray) {
     metadataArray.forEach(function(metadata) {
       let time = metadata.cueTime + this.timestampOffset;
+      let videoDuration = Number.isFinite(this.mediaSource_.duration) ?
```
imbcmdth: I wonder if Babel will polyfill `Number.isFinite` in older browsers? This could also be written as `let videoDuration = Math.min(this.mediaSource_.duration, Number.MAX_VALUE);`.

imbcmdth:
I wrote this up yesterday. At the end it has a pseudo algorithm for generating cues that might be interesting. I am not saying this is the only way to do it but it might help to see what I was thinking: https://github.com/videojs/videojs-contrib-hls/issues/830

src/add-text-track-data.js
```diff
@@ -64,21 +64,54 @@ const addTextTrackData = function(sourceHandler, captionArray, metadataArray) {
...
+     * the endTime of each cue is the startTime of the next cue
+     * the endTime of last cue is the duration of the video
+     */
+    if (sourceHandler.metadataTrack_ && sourceHandler.metadataTrack_.cues) {
+      for (let i = 0; i <= sourceHandler.metadataTrack_.cues.length - 1; i++) {
+        if (i === sourceHandler.metadataTrack_.cues.length - 1) {
+          sourceHandler.metadataTrack_.cues[i].endTime = videoDuration;
+        } else if (sourceHandler.metadataTrack_.cues[i].endTime !==
+          sourceHandler.metadataTrack_.cues[i + 1].startTime &&
+          i < sourceHandler.metadataTrack_.cues.length - 1) {
```
forbesjo: This check is unnecessary.

src/add-text-track-data.js
```diff
@@ -99,11 +99,9 @@ const addTextTrackData = function(sourceHandler, captionArray, metadataArray) {
       for (let i = 0; i <= sourceHandler.metadataTrack_.cues.length - 1; i++) {
```
imbcmdth: This is better written as `for (let i = 0; i < sourceHandler.metadataTrack_.cues.length; i++) {`. Most developers are used to this and will automatically read it as "for each element in the array".

src/add-text-track-data.js
```diff
@@ -99,11 +99,9 @@ const addTextTrackData = function(sourceHandler, captionArray, metadataArray) {
...
         if (i === sourceHandler.metadataTrack_.cues.length - 1) {
           sourceHandler.metadataTrack_.cues[i].endTime = videoDuration;
-        } else if (sourceHandler.metadataTrack_.cues[i].endTime !==
-          sourceHandler.metadataTrack_.cues[i + 1].startTime &&
-          i < sourceHandler.metadataTrack_.cues.length - 1) {
+        } else {
           sourceHandler.metadataTrack_.cues[i].endTime =
-            sourceHandler.metadataTrack_.cues[i + 1].startTime;
+          sourceHandler.metadataTrack_.cues[i + 1].startTime;
```
imbcmdth: Looking at this function in it's entirety I think it might help to remove the odd condition from the loop and change the whole thing into:

for (let i = 0; i <= sourceHandler.metadataTrack_.cues.length - 1; i++) {
@imbcmdth     Video.js member
imbcmdth added a line comment 2 minutes ago
Usually, this is better written as for (let i = 0; i < sourceHandler.metadataTrack_.cues.length; i++) {. Most developers are used to this and will automatically read it as "for each element in the array".

EDIT:

Actually looking at the whole for loop I think it makes more sense to do what I talk about in the comment below...

```javascript
let cues = sourceHandler.metadataTrack_.cues;

for (let i = 0; i <cues.length - 1; i++) {
      cues[i].endTime = cues[i + 1].startTime;
}

cues[cues.length - 1].endTime = videoDuration;
```

src/add-text-track-data.js
```diff
@@ -64,21 +64,49 @@ const addTextTrackData = function(sourceHandler, captionArray, metadataArray) {
   }
 
   if (metadataArray) {
+    let videoDuration;
+
+    if (isNaN(sourceHandler.mediaSource_.duration) ||
+      sourceHandler.mediaSource_.duration === Infinity ||
+      sourceHandler.mediaSource_.duration === -Infinity) {
```
gesinger: I wonder if it is safe to use `Math.abs(sourceHandler.mediaSource_.duration)` instead of checking for both Infinity and -Infinity.

gesinger:
I think if possible it would be good to add a test for the case where 'sourceended' is fired, but otherwise looks pretty good!

src/add-text-track-data.js
```diff
@@ -64,21 +64,48 @@ const addTextTrackData = function(sourceHandler, captionArray, metadataArray) {
...
+     */
+    if (sourceHandler.metadataTrack_ && sourceHandler.metadataTrack_.cues) {
+      let cues = sourceHandler.metadataTrack_.cues;
+
+      for (let i = 0; i < cues.length - 1; i++) {
+        cues[i].endTime = cues[i + 1].startTime;
+      }
+      cues[cues.length - 1].endTime = videoDuration;
+
+      sourceHandler.mediaSource_.on('sourceended', (event) => {
```
forbesjo: This is being registered for every addTextTrackData call

src/add-text-track-data.js
```diff
@@ -64,21 +74,46 @@ const addTextTrackData = function(sourceHandler, captionArray, metadataArray) {
...
+     * the endTime of last cue is the duration of the video
+     */
+    if (sourceHandler.metadataTrack_ && sourceHandler.metadataTrack_.cues) {
+      let cues = sourceHandler.metadataTrack_.cues;
+      let cuesArray = [];
+
+      for (let j = 0; j < cues.length; j++) {
+        cuesArray.push(cues[j]);
+      }
+      cuesArray.sort(function(first, second) {
```
forbesjo: You can use an [arrow function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions) here.

forbesjo:
LGTM

src/add-text-track-data.js
```diff
@@ -40,6 +40,16 @@ const deprecateOldCue = function(cue) {
   });
 };
 
+const durationOfVideo = function(duration) {
```
mjneil: I think it would be a good idea to export this function and import it in html-media-source.js since you repeat the same logic

src/html-media-source.js
```diff
@@ -180,6 +180,24 @@ export default class HtmlMediaSource extends videojs.EventTarget {
...
 
+    this.on('sourceended', (event) => {
+      let duration = this.duration;
+
+      for (let i = 0; i < this.sourceBuffers.length; i++) {
+        let sourcebuffer = this.sourceBuffers[i];
+        let cues = sourcebuffer.metadataTrack_.cues;
+
+        if (cues) {
+          if (isNaN(duration) || Math.abs(duration) === Infinity) {
```
mjneil: The duration logic should be done moved to before the for loop since duration won't change with each duration. Also could just call the function if you export from add-text-track-data.js as mentioned in my other comment.

mjneil:
lgtm

PR #101, opened 2016-09-06
Disabled source buffers may still have buffered content

dmlap:
Don't hide buffered content in a source buffer with audio disabled because well-behaved clients could misinterpret this. For example, a client may choose not to remove buffered data from a disabled audio source buffer since it appears empty.


gesinger:
Might be worth adding a test for the behavior of the buffered property with audioDisabled_ set to true/false, and a test for remove that shows audio buffer is removed with audioDisabled_ set to true/false. Otherwise, LGTM.

dmlap:
@gesinger added a test for removal with audio disabled and some comments to clarify that was being tested. There are already existing tests that check buffered and remove without audio disabled.

gesinger:
LGTM

PR #103, opened 2016-09-16
Added browserify/webpack tests

ruchigupta19:


mjneil:
It seems like `tmp/browserify.test.js` and `tmp/webpack.test.js` are never being created. for `browserify.test.js` this doesn't really seem like an issue since the test is being added to `dist-test/videojs-contrib-media-sources.js` in `scripts/build-test.js` using browserify, but `webpack.test.js` is being added also using browserify and not webpack. Do we want to create another step in `build-test.js` that builds `webpack.test.js` using webpack, or just remove the inclusion of the tmp folder scripts?

mjneil:
I think it would be a good idea to add `watch:test:browserify` and `watch:test:webpack` scripts as well since build isnt run when using `npm start` or `npm watch`

forbesjo:
LGTM

PR #104, opened 2016-09-22
append init segment data on audio track changes

mjneil:
required for [pull request](https://github.com/videojs/mux.js/pull/120)

### DIFF
--- a/src/html-media-source.js
+++ b/src/html-media-source.js
@@ -132,6 +132,8 @@ export default class HtmlMediaSource extends videojs.EventTarget {
         //      what stream is the video stream, rather than relying on videoTracks
         /* eslinst-enable */
 
+        sourceBuffer.appendInitSegment_ = true;
+
         if (sourceBuffer.videoCodec_ && sourceBuffer.audioCodec_) {
           // combined
           sourceBuffer.audioDisabled_ = combined;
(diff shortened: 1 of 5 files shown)


src/html-media-source.js
```diff
@@ -131,6 +131,8 @@ export default class HtmlMediaSource extends videojs.EventTarget {
         //      what stream is the video stream, rather than relying on videoTracks
         /* eslinst-enable */
 
+        sourceBuffer.appendInitSegment_ = true;
```
gesinger: Do we have to do this on all of 'change', 'addtrack', and 'removetrack'?

gesinger:
Looks pretty good, but might be good to add some unit tests for when appendInitSegment_ is changed and that processPendingSegments_ adds the init segments to video and audio segment lists when appendInitSegment_ is true (and doesn't when false).

test/html.test.js
```diff
@@ -438,6 +443,75 @@ function() {
   QUnit.equal(mp4Segments[0][1], 6, 'fragment contains the correct second byte');
 });
 
+QUnit.test(
+'only appends init segment for first segment or on audio changes',
```
gesinger: I know that audio changes change the appendInitSegment_ value, but I didn't see anything in the test that identifies audio changes.

mjneil: You're right. I've updated the test to trigger events on the player.

gesinger:
LGTM

imbcmdth:
LGTM

PR #105, opened 2016-10-03
remove window from transmuxer-worker

sh84:
If you use video.js+videojs-contrib-hls and build them using webpack, than need to use webworkify-webpack package.
webworkify-webpack make global variable window = {} for workers.
And src/transmuxer-worker.js used window.postMessage..
So, in this case it does not work with the error window.postMessage is not a function.
But window variable not needed in workers, and if delete it - all work correctly.


forbesjo:
I believe this was resolved with using web-wackify

PR #106, opened 2016-10-18
Only trim FLV tags when seeking to prevent triming I frames

mjneil:
The flash source buffer trims any FLV tags that are before the end of the current buffer to prevent overlapping or misalignment on appends. This was causing some I frames to be trimmed in some streams, which would create blocky artifacts during playback.

This change removes the trimming of tags before the end of the current buffer and also uses the tech's seeking property to determine when a seek happens instead of inferring a seek based on the state of the buffer.

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
   ],
   "dependencies": {
     "global": "^4.3.0",
-    "mux.js": "^3.0.0",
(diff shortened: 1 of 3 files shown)


dmlap:
LGTM

forbesjo:
Could you bump up the mux.js version when it's released?

PR #107, opened 2016-10-18
change this.mediaSource to this.mediaSource_ in flash source buffer

forbesjo:
This https://github.com/videojs/videojs-contrib-media-sources/blob/master/src/add-text-track-data.js#L77 does not work in flash mode


mjneil:
LGTM

PR #109, opened 2016-10-21
append init segment to video buffer for every segment

mjneil:
Some content may be less than ideal and change the video info on segment boundaries in the same rendition. In firefox, this causes rendering issues (green artifacts) since the init segment should change on a video info change.

This change will append an init segment to the video buffer every segment, but still only append init segments for audio on media or audio info change to prevent the audio pop experienced in firefox on segment boundaries.

### DIFF
--- a/src/virtual-source-buffer.js
+++ b/src/virtual-source-buffer.js
@@ -33,7 +33,7 @@ export default class VirtualSourceBuffer extends videojs.EventTarget {
     this.audioCodec_ = null;
     this.videoCodec_ = null;
     this.audioDisabled_ = false;
-    this.appendInitSegment_ = true;
+    this.appendAudioInitSegment_ = true;
 
     let options = {
       remux: false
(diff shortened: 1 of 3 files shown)


imbcmdth:
LGTM

src/virtual-source-buffer.js
```diff
@@ -437,10 +437,6 @@ export default class VirtualSourceBuffer extends videojs.EventTarget {
     }
 
     if (this.appendInitSegment_) {
```
gesinger: Maybe we can handle it in a different PR, but we should probably change this to appendAudioInitSegment_ now.

gesinger: Same for naming of the 'only appends init segment for first segment or on audio/media changes' test

gesinger:
Might also be good to have a test where we create the video data message to verify this behavior.

src/virtual-source-buffer.js
mjneil: done

gesinger:
We should probably test that no matter whether appendAudioInitSegment_ is true or false, the video segment is still appended.

PR #113, opened 2016-11-03
Metadata cues may have length zero

dmlap:
It's not safe to assume that there is at least one element in metadata cues when addTextTrackData() is invoked. Double check the array is there and non-empty before tweaking start and end times. Add tests for that module.


forbesjo:
The same check should be added to https://github.com/videojs/videojs-contrib-media-sources/blob/master/src/html-media-source.js#L201

mjneil:
LGTM once @forbesjo suggestion is added

ruchigupta19:
LGTM as soon as @forbesjo suggestion is acknowledged

PR #116, opened 2016-11-10
catch error on appendbuffer

mjneil:
appendBuffer can throw a QuotaExceededError. This would cause the player to fail without videojs reporting any errors. This will report the error to the player so the error actually gets reported.


imbcmdth:
LGTM

PR #118, opened 2016-11-14
Start using remote TextTracks because they can be properly removed

imbcmdth:
We were using `videojs.addTextTrack` which creates `TextTrack` objects in memory and associates them with a video element. These are not removable from the video element once added. This PR switches to `videojs.addRemoteTextTrack`and then carefully removes them whenever the MediaSource is closed (in the Html5 case) or the HLS sourceHandler is disposed of (in the Flash case).

### DIFF
--- a/src/create-text-tracks-if-necessary.js
+++ b/src/create-text-tracks-if-necessary.js
@@ -1,6 +1,7 @@
 /**
  * @file create-text-tracks-if-necessary.js
  */
+import {removeExistingTrack} from './cleanup-text-tracks';
 
 /**
  * Create text tracks on video.js if they exist on a segment.
@@ -11,18 +12,27 @@
  * @private
  */
 const createTextTracksIfNecessary = function(sourceBuffer, mediaSource, segment) {
+  const player = mediaSource.player_;
+
   // create an in-band caption track if one is present in the segment
   if (segment.captions &&
       segment.captions.length &&
       !sourceBuffer.inbandTextTrack_) {
-    sourceBuffer.inbandTextTrack_ = mediaSource.player_.addTextTrack('captions', 'cc1');
+    removeExistingTrack(player, 'captions', 'cc1');
+    sourceBuffer.inbandTextTrack_ = player.addRemoteTextTrack({
+      kind: 'captions',
+      label: 'cc1'
+    }, false).track;
   }
 
   if (segment.metadata &&
       segment.metadata.length &&
       !sourceBuffer.metadataTrack_) {
-    sourceBuffer.metadataTrack_ =
-      mediaSource.player_.addTextTrack('metadata', 'Timed Metadata');
+    removeExistingTrack(player, 'metadata', 'Timed Metadata', true);
+    sourceBuffer.metadataTrack_ = player.addRemoteTextTrack({
+      kind: 'metadata',
+      label: 'Timed Metadata'
+    }, false).track;
     sourceBuffer.metadataTrack_.inBandMetadataTrackDispatchType =
       segment.metadata.dispatchType;
   }
--- a/src/flash-media-source.js
+++ b/src/flash-media-source.js
@@ -6,6 +6,7 @@ import videojs from 'video.js';
 import FlashSourceBuffer from './flash-source-buffer';
 import FlashConstants from './flash-constants';
 import {parseContentType} from './codec-utils';
+import {cleanupTextTracks} from './cleanup-text-tracks';
 
 /**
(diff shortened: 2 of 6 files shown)


src/flash-media-source.js
```diff
@@ -36,6 +37,12 @@ export default class FlashMediaSource extends videojs.EventTarget {
         }
       });
 
+      if (this.tech_.hls) {
+        this.tech_.hls.on('dispose', () => {
```
gesinger: I know that it looks like video.js' tech doesn't trigger dispose, but maybe we should consider adding it later, so we don't have to check for HLS specifically (or listening to a different type of event).

test/flash.test.js
```diff
@@ -739,3 +740,64 @@ QUnit.test('fires loadedmetadata after first segment append', function() {
...
+    tags: {
+      videoTags: [],
+      audioTags: []
+    },
+    metadata,
+    captions
+  });
+  sourceBuffer.segmentParser_.trigger('done');
+
+  QUnit.equal(addedTracks.length, 2, 'created two remote TextTracks');
```
gesinger: We should check which ones were created.

test/flash.test.js
```diff
@@ -739,3 +740,64 @@ QUnit.test('fires loadedmetadata after first segment append', function() {
...
+  this.player.tech_.hls.trigger('dispose');
+  QUnit.equal(removedTracks.length, 2, 'removed both TextTracks');
```
gesinger: Same for checking which ones removed.

test/html.test.js
```diff
@@ -1073,6 +1086,69 @@ QUnit.test('translates metadata events into WebVTT cues', function() {
...
+    removeRemoteTextTrack(track) {
+      removedTracks.push(track.kind);
+    }
+  };
+  sourceBuffer.timestampOffset = 10;
+
+  sourceBuffer.transmuxer_.onmessage(createDataMessage('video', new Uint8Array(1), {
+    metadata
+  }));
+  sourceBuffer.transmuxer_.onmessage(createDataMessage('video', new Uint8Array(1), {
```
gesinger: Should we check in-between these two messages?

src/create-text-tracks-if-necessary.js
```diff
@@ -11,18 +12,27 @@
  * @private
  */
 const createTextTracksIfNecessary = function(sourceBuffer, mediaSource, segment) {
```
gesinger: May be worth adding tests for this.

src/flash-media-source.js
imbcmdth: Techs do have a dispose but they are only dispose when switching techs (html->flash) but sourceHandlers have a dispose too and they trigger it whenever the source changes (their dispose gets triggered via HLSHandler's superclass in `Component#dispose`).

So while I agree that this line is kind of cludgy, it would be necessary since video.js will properly clean up remote text tracks whenever the source changes. This is just insurance that they get cleaned up even if using an older version of video.js.

test/flash.test.js
imbcmdth: We have tests that catch this scenario already and make sure the proper tracks are created for captions and metadata individually but it is a good idea to check that here for completeness.

test/flash.test.js
imbcmdth: We have tests that catch this scenario already and make sure the proper tracks are created for captions and metadata individually but it is a good idea to check that here for completeness.

test/html.test.js
imbcmdth: Nah, in fact I should just collapse the two messages into one. No reason to keep both. Thanks for pointing that out.

test/html.test.js
```diff
@@ -1073,6 +1086,147 @@ QUnit.test('translates metadata events into WebVTT cues', function() {
...
+
+  QUnit.equal(removedTracks.length, 2, 'removed two text tracks');
+  QUnit.equal(removedTracks.filter(t => ['captions', 'metadata'].indexOf(t.kind) === -1).length,
+              0,
+              'removed only the expected two remote TextTracks');
+
+  QUnit.equal(addedTracks.length, 2, 'created two text tracks');
+  QUnit.equal(addedTracks.filter(t => ['captions', 'metadata'].indexOf(t.kind) === -1).length,
+              0,
+              'created only the expected two remote TextTracks');
```
gesinger: Maybe it's overboard, since the remove should be fine, but it might be worth checking to verify that the newly created tracks are the new ones.

PR #119, opened 2016-11-17
Replaced the adaptive chunking code for Flash with a constant one with a bandwidth of roughly 1MB/s

imbcmdth:
The purpose of this code was to upload to Flash as quickly as possible. The intent was noble but it never quite managed to work without introducing jankiness at higher bitrates.

### DIFF
--- a/src/flash-constants.js
+++ b/src/flash-constants.js
@@ -4,36 +4,17 @@
 /**
  * The maximum size in bytes for append operations to the video.js
  * SWF. Calling through to Flash blocks and can be expensive so
- * tuning this parameter may improve playback on slower
- * systems. There are two factors to consider:
- * - Each interaction with the SWF must be quick or you risk dropping
- * video frames. To maintain 60fps for the rest of the page, each append
- * must not  take longer than 16ms. Given the likelihood that the page
- * will be executing more javascript than just playback, you probably
- * want to aim for less than 8ms. We aim for just 4ms.
- * - Bigger appends significantly increase throughput. The total number of
- * bytes over time delivered to the SWF must exceed the video bitrate or
- * playback will stall.
- *
- * We adaptively tune the size of appends to give the best throughput
- * possible given the performance of the system. To do that we try to append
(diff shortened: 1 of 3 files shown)


imbcmdth:
Required for https://github.com/videojs/videojs-contrib-hls/pull/918

src/flash-constants.js
```diff
@@ -4,36 +4,17 @@
...
- * as much as possible in TIME_PER_TICK and while tuning the size of appends
- * dynamically so that we only append about 4-times in that 4ms span.
- *
- * The reason we try to keep the number of appends around four is due to
- * externalities such as Flash load and garbage collection that are highly
- * variable and having 4 iterations allows us to exit the loop early if
- * an iteration takes longer than expected.
+ * we chunk data and pass through 4KB at a time, yielding to the
+ * browser between chunks. This gives a theoretical maximum rate of
+ * 1MB/s into Flash. Any higher and we being to drop frames and UI
```
gesinger: s/being/begin

src/flash-source-buffer.js
```diff
@@ -260,58 +252,35 @@ export default class FlashSourceBuffer extends videojs.EventTarget {
...
     }
-
-    // We also make sure that the chunk-size doesn't drop below 1KB or
-    // go above 1MB as a sanity check
-    this.chunkSize_ = Math.max(
-      FlashConstants.MIN_CHUNK,
-      Math.min(this.chunkSize_, FlashConstants.MAX_CHUNK));
+    let b64str = window.btoa(binary);
+
+    // bypass normal ExternalInterface calls and pass xml directly
```
gesinger: That's crazy to me that this is faster

mjneil:
LGTM

src/flash-source-buffer.js
imbcmdth: Yeah... it was back in the day. I don't know if the performance of ExternalInterface calls have improved in recent times.

gesinger:
LGTM

PR #120, opened 2016-11-25
Don't crash if the initSegment isn't available.

arendjr:
This can happen when using the HLS plugin and we switch from one stream to another due to fluctuations in available bandwidth.

### DIFF
--- a/src/virtual-source-buffer.js
+++ b/src/virtual-source-buffer.js
@@ -446,8 +446,10 @@ export default class VirtualSourceBuffer extends videojs.EventTarget {
 
     // Merge multiple video and audio segments into one and append
     if (this.videoBuffer_) {
-      sortedSegments.video.segments.unshift(sortedSegments.video.initSegment);
-      sortedSegments.video.bytes += sortedSegments.video.initSegment.byteLength;
+      if (sortedSegments.video.initSegment) {
+        sortedSegments.video.segments.unshift(sortedSegments.video.initSegment);
+        sortedSegments.video.bytes += sortedSegments.video.initSegment.byteLength;
+      }
       this.concatAndAppendSegments_(sortedSegments.video, this.videoBuffer_);
       // TODO: are video tracks the only ones with text tracks?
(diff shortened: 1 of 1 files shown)


mjneil:
Do you have a reproducible test case that exhibits the crash? While this PR probably wouldn't hurt any to add, initSegment not being available may just be a symptom of another problem, so I think it would be good to be able to investigate further to see if we can figure out the root cause of the problem and fix it there.

gesinger:
Adding onto what @mjneil said, https://github.com/videojs/videojs-contrib-hls/pull/939 addresses a problem where HLS updates a playlist and a segment that is currently being processed is no longer on the playlist (this is mainly in the case of live scenarios, where the playlist update is due to a refresh, and the segment falls off the end of it). In this case, we reference a missing segment, pass empty data to the transmuxer, and media sources gets empty data back, which may cause this issue.

Just wanted to check to see if this seems to be the case you are seeing or your case comes from a different scenario.

And thanks for contributing!

arendjr:
If I remember correctly, I could pretty reliably reproduce it by using the bandwidth throttling feature in Chrome so it would force the player to switch bandwidth (though I'm not sure anymore whether the crash occurred when it switched to a lower bandwidth version or later when it switched to a higher bandwidth stream as I unthrottled it).

I can tell you none of the streams I was testing with were live streams, but all sources were prerendered MP4 files served through HLS by Wowza.

gaxunil:
We are seeing a very similar issue to this but on the audio instead:
https://github.com/videojs/videojs-contrib-media-sources/issues/181
We are in a live, multiple players on a page scenario and maybe are hitting the same conditions that was triggering this issue.  We can't yet reproduce it at will, usually happens when bandwidth starts to become limited and the streams shift.

forbesjo:
Thank you for your PR but we will not be accepting new changes for this repo and will be [archived](https://help.github.com/articles/about-archiving-repositories/) very soon. I would advise you to open your PR against the next iteration of videojs-contrib-hls at https://github.com/videojs/http-streaming.

PR #121, opened 2016-12-21
Fixed metadata cue mapping so that it considers groups cues with the same startTime and remaps them collectively to the same endTime

imbcmdth:
When there are ID3 tags with multiple frames, this change will create several overlapping cues that all start at the same time and all end at the same time.

This behavior is somewhat different from what Safari does but it makes more sense. I think Safari is "doing it wrong" anyway since it creates some cues that are literally 1/90000th of a second long when there are ID3 tags with multiple frames.

### DIFF
--- a/src/add-text-track-data.js
+++ b/src/add-text-track-data.js
@@ -127,7 +127,7 @@ const addTextTrackData = function(sourceHandler, captionArray, metadataArray) {
       // Map each cue group's endTime to the next group's startTime
       sortedStartTimes.forEach((startTime, idx) => {
         let cueGroup = cuesGroupedByStartTime[startTime];
-        let nextTime = sortedStartTimes[idx + 1] || videoDuration;
+        let nextTime = Number(sortedStartTimes[idx + 1]) || videoDuration;
 
         // Map each cue's endTime the next group's startTime
         cueGroup.forEach((cue) => {
--- a/src/add-text-track-data.js
+++ b/src/add-text-track-data.js
@@ -102,17 +102,38 @@ const addTextTrackData = function(sourceHandler, captionArray, metadataArray) {
       let cues = sourceHandler.metadataTrack_.cues;
       let cuesArray = [];
 
-      for (let j = 0; j < cues.length; j++) {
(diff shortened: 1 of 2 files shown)


src/add-text-track-data.js
```diff
@@ -102,17 +102,38 @@ const addTextTrackData = function(sourceHandler, captionArray, metadataArray) {
...
+                                   .sort((a, b) => Number(a) - Number(b));
+
+      // Map each cue group's endTime to the next group's startTime
+      sortedStartTimes.forEach((startTime, idx) => {
+        let cueGroup = cuesGroupedByStartTime[startTime];
+        let nextTime = sortedStartTimes[idx + 1] || videoDuration;
+
+        // Map each cue's endTime the next group's startTime
+        cueGroup.forEach((cue) => {
+          cue.endTime = nextTime;
```
mjneil: nextTime will be a string here instead of a number if it came from `sortedStartTimes[idx + 1]`

src/add-text-track-data.js
```diff
@@ -102,17 +102,38 @@ const addTextTrackData = function(sourceHandler, captionArray, metadataArray) {
...
+      // ...disregarding cues with a falsey value
+      for (let i = 0; i < cues.length; i++) {
+        if (cues[i]) {
+          cuesArray.push(cues[i]);
         }
       }
-      cuesArray[cuesArray.length - 1].endTime = videoDuration;
+
+      // Group cues by their startTime value
+      let cuesGroupedByStartTime = cuesArray.reduce((obj, cue) => {
```
mjneil: We could avoid having to need this `reduce` call completely by creating the `cuesGroupedByStartTime` object above in the for loop instead of populating the `cuesArray` that is only used for this reduce call.

imbcmdth: We can but I am not sure making the for loop more complicated is a good choice - it seems like an early optimization.

I like the step-by-step approach that is easier to reason about in the long term because I have a bad memory!

mjneil:
LGTM

PR #122, opened 2016-12-22
Move translateLegacyCodecs into codec-utils

gesinger:


mjneil:
LGTM

test/codec-utils.test.js
```diff
@@ -0,0 +1,28 @@
+import Qunit from 'qunit';
+import { translateLegacyCodecs } from '../src/codec-utils';
+
+const { deepEqual, module, test } = Qunit;
```
mjneil: As awesome and clean as I think this is, we don't do this any of the other tests and if we make the jump to QUnit 2.0 in this project, it might make the transition slightly more annoying

gesinger: I actually just copied it from https://github.com/videojs/videojs-contrib-media-sources/blob/master/test/add-text-track-data.test.js , but I can change it back.

mjneil: Well, I stand corrected, since it's already in one place, I don't have as strong a preference for one way or the other

forbesjo:
Looks good to me

PR #123, opened 2017-01-03
Set audio append start in transmuxer

gesinger:
Used to activate https://github.com/videojs/mux.js/pull/130

### DIFF
--- a/src/transmuxer-worker.js
+++ b/src/transmuxer-worker.js
@@ -119,6 +119,10 @@ class MessageHandlers {
     this.transmuxer.setBaseMediaDecodeTime(Math.round(timestampOffset * 90000));
   }
 
+  setAudioAppendStart(data) {
+    this.transmuxer.setAudioAppendStart(Math.round(data.appendStart * 90000));
+  }
+
   /**
    * Forces the pipeline to finish processing the last segment and emit it's
    * results.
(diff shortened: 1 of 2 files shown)


src/transmuxer-worker.js
```diff
@@ -119,6 +119,10 @@ class MessageHandlers {
     this.transmuxer.setBaseMediaDecodeTime(Math.round(timestampOffset * 90000));
   }
 
+  setAudioAppendStart(data) {
+    this.transmuxer.setAudioAppendStart(Math.round(data.appendStart * 90000));
```
imbcmdth: Should this be `Math.ceil`?? Since we take so much care elsewhere to make sure we don't overlap existing data, the possibility of this one rounding down the buffer seems a little worrying.

src/virtual-source-buffer.js
```diff
@@ -325,6 +325,15 @@ export default class VirtualSourceBuffer extends videojs.EventTarget {
     // Start the internal "updating" state
     this.bufferUpdating_ = true;
 
+    if (this.audioBuffer_ && this.audioBuffer_.buffered.length) {
+      let audioBuffered = this.audioBuffer_.buffered;
+
+      this.transmuxer_.postMessage({
+        action: 'setAudioAppendStart',
+        appendStart: audioBuffered.end(audioBuffered.length - 1)
```
imbcmdth: What if the `audioBuffered` length is 0?

gesinger: Checked for 0 above (in conditional)

imbcmdth: OK

imbcmdth:
LGTM

PR #124, opened 2017-01-05
drop audio tags that come before buffered end

mjneil:
PR #106 removed the dropping of FLV tags that came before the end of the buffer before appending to prevent artifacts due to missing I frames. This created an issue that caused audio to replay when appending a repeated segment from another rendition. It appears that when we append video data that has timestamps before the playhead, flash ignores those tags and does not replay video, but seems to not take into account the timestamps of the audio and appends the audio data to the buffer, causing a stall in video playback and a repeat in audio. 

This change trims the audio and video FLV tags separately to bring back dropping of audio tags that come before the end of the buffer, but keeping the video tags. Both types are still dropped on seeks.

### DIFF
--- a/src/flash-constants.js
+++ b/src/flash-constants.js
@@ -13,8 +13,8 @@
  */
 const flashConstants = {
   // times in milliseconds
-  TIME_BETWEEN_CHUNKS: 4,
-  BYTES_PER_CHUNK: 4096
+  TIME_BETWEEN_CHUNKS: 2,
+  BYTES_PER_CHUNK: 4096 * 8
 };
 
 export default flashConstants;
--- a/src/flash-constants.js
+++ b/src/flash-constants.js
@@ -13,7 +13,7 @@
  */
 const flashConstants = {
   // times in milliseconds
-  TIME_BETWEEN_CHUNKS: 2,
+  TIME_BETWEEN_CHUNKS: 4,
   BYTES_PER_CHUNK: 4096
 };
 
--- a/src/flash-constants.js
+++ b/src/flash-constants.js
@@ -14,7 +14,7 @@
 const flashConstants = {
   // times in milliseconds
   TIME_BETWEEN_CHUNKS: 2,
-  BYTES_PER_CHUNK: 4096 * 8
+  BYTES_PER_CHUNK: 4096
 };
 
 export default flashConstants;
--- a/src/flash-source-buffer.js
+++ b/src/flash-source-buffer.js
@@ -304,10 +304,7 @@ export default class FlashSourceBuffer extends videojs.EventTarget {
   convertTagsToData_(segmentData) {
     let segmentByteLength = 0;
     let tech = this.mediaSource_.tech_;
-    let targetVideoPts = 0;
-    let targetAudioPts = 0;
-    let i;
-    let j;
+    let targetPts = 0;
     let segment;
     let filteredAudioTags = [];
     let filteredVideoTags = [];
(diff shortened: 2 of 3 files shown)


src/flash-source-buffer.js
```diff
@@ -304,47 +304,68 @@ export default class FlashSourceBuffer extends videojs.EventTarget {
...
+    targetAudioPts *= 1e3;
+    targetAudioPts += this.basePtsOffset_;
 
     // skip tags with a presentation time less than the seek target
-    for (i = 0; i < tags.length; i++) {
-      if (tags[i].pts >= targetPts) {
-        filteredTags.push(tags[i]);
+    for (i = 0; i < videoTags.length; i++) {
+      if (videoTags[i].pts >= targetVideoPts) {
+        filteredVideoTags.push(videoTags[i]);
```
imbcmdth: We seem to not culling any video tags at all (effectively). I think the idea should cull with a bit of awareness of the GOP structure and cull entire GOPS that precede the `targetVideoPts` (which would be the seek time or the end of the buffer). Doing this means playback starts within complete GOPS and should eliminate the "smearing" that was experienced previously without pushing too many extra frames into Flash.

test/flash.test.js
```diff
@@ -343,6 +343,54 @@ QUnit.test('drops tags before currentTime when seeking', function() {
...
+
+  endTime = 10 + 7;
+  this.mediaSource.tech_.trigger('seeking');
+  sourceBuffer.appendBuffer(new Uint8Array(10));
+  this.swfCalls.length = 0;
+  timers.runAll();
+
+  // 0 - 9 for video tags not droped, 7 - 9 for audio tags (0 - 6 dropped for being before buffered end)
+  QUnit.deepEqual(this.swfCalls[0].arguments[0], [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 7, 8, 9],
+            'three tags are appended');
```
imbcmdth: This assertions description doesn't make sense - 13 tags are inserted!

test/flash.test.js
```diff
@@ -343,6 +343,54 @@ QUnit.test('drops tags before currentTime when seeking', function() {
...
+  // starting PTS value
+  while (i--) {
+    tags_.unshift(
+      makeFlvTag((i * 1000) + (29 * 1000),
+        new Uint8Array([i])));
+  }
+  sourceBuffer.segmentParser_.trigger('data', {
+    tags: {
+      videoTags: tags_,
+      audioTags: tags_
```
imbcmdth: `videoTags` and `audioTags` should have easily identifiable tags so that we can be sure that the right ones are culled. Right now you can't be sure which 0-9 and which 7-9 there are represented in the assertion below.

test/flash.test.js
```diff
@@ -343,6 +343,54 @@ QUnit.test('drops tags before currentTime when seeking', function() {
...
+  });
+
+  endTime = 10 + 7;
+  this.mediaSource.tech_.trigger('seeking');
+  sourceBuffer.appendBuffer(new Uint8Array(10));
+  this.swfCalls.length = 0;
+  timers.runAll();
+
+  // 0 - 9 for video tags not droped, 7 - 9 for audio tags (0 - 6 dropped for being before buffered end)
+  QUnit.deepEqual(this.swfCalls[0].arguments[0], [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 7, 8, 9],
```
imbcmdth: Shouldn't the video and audio tags be interleaved by time so that the output would be: `[0, 1, 2, 3, 4, 5, 6, 7, 7, 8, 8, 9, 9]`??

src/flash-constants.js
```diff
@@ -13,7 +13,7 @@
  */
 const flashConstants = {
   // times in milliseconds
-  TIME_BETWEEN_CHUNKS: 4,
+  TIME_BETWEEN_CHUNKS: 2,
```
imbcmdth: Please revert this change, please. Other than this it LGTM!

forbesjo:
Looks good to me.

PR #126, opened 2017-01-27
Flash Performance Enhancements

mjneil:
Currently, the flash fallback is not so great. In situations where a user has low bandwidth or a not so powerful cpu, they may experience frequent and long buffering through playback.  This is primarily due to the way segments are transferred to the swf.  In order to not freeze up the main thread, we have to chunk up the segment data into very small 4k chunks to transmux and append each one asynchronously. 

**This PR requires [video-js-swf #218](https://github.com/videojs/video-js-swf/pull/218) and [mux.js #139](https://github.com/videojs/mux.js/pull/139)**

This PR brings two main enhancements.
* Transmuxing in a web worker
  * By moving the transmuxer to the webworker, we can transmux the entire segment at once without taking up the main thread. This gives us more "time" to process things in the main thread.
* Pull from flash instead of Push to flash
  * Instead of pushing data to flash like we do now with the ExternalInterface CallFunction, this adds an intermediate call to the swf that tells flash we have data ready for it. Flash then uses ExternalInterface to call a global function in javascript, which returns the segment bytes. Allowing the swf to initiate and handle the transfer seems to allow larger chunks of transfer without slowing the javascript thread. This change allows us to increase the chunksize 8-fold to 32kb.

### DIFF
--- a/src/flash-source-buffer.js
+++ b/src/flash-source-buffer.js
@@ -87,7 +87,12 @@ export default class FlashSourceBuffer extends videojs.EventTarget {
         )
       )
     );
-    this.mediaSource_.swfObj.vjs_appendBuffer(encodedHeader);
+
+    window.encodedHeaderSuperSecret = function () {
+      throw encodedHeader;
+    };
+
+    this.mediaSource_.swfObj.vjs_appendBuffer('encodedHeaderSuperSecret');
 
     // TS to FLV transmuxer
     this.transmuxer_ = work(transmuxWorker);
@@ -163,7 +168,6 @@ export default class FlashSourceBuffer extends videojs.EventTarget {
       error.code = 11;
       throw error;
     }
-
     this.updating = true;
     this.mediaSource_.readyState = 'open';
     this.trigger({ type: 'update' });
@@ -267,30 +271,29 @@ export default class FlashSourceBuffer extends videojs.EventTarget {
     this.bufferSize_ -= chunk.byteLength;
 
     // base64 encode the bytes
-    let binary = '';
+    let binary = [];
     let length = chunk.byteLength;
 
     for (let i = 0; i < length; i++) {
-      binary += String.fromCharCode(chunk[i]);
+      binary.push(String.fromCharCode(chunk[i]));
     }
-    let b64str = window.btoa(binary);
+    let b64str = window.btoa(binary.join(''));
+
+    window.throwDataSuperSecret = function () {
+      // schedule another append if necessary
+      if (this.bufferSize_ !== 0) {
+        scheduleTick(this.processBuffer_.bind(this));
+      } else {
+        this.updating = false;
+        this.trigger({ type: 'updateend' });
+      }
+
+      throw b64str;
+    }.bind(this);
 
     // bypass normal ExternalInterface calls and pass xml directly
     // IE can be slow by default
-    this.mediaSource_.swfObj.CallFunction(
-      '<invoke name="vjs_appendBuffer"' +
-      'returntype="javascript"><arguments><string>' +
-      b64str +
-      '</string></arguments></invoke>');
-
-    // schedule another append if necessary
-    if (this.bufferSize_ !== 0) {
-      scheduleTick(this.processBuffer_.bind(this));
-    } else {
-      this.updating = false;
-      this.trigger({ type: 'updateend' });
-
-    }
+    this.mediaSource_.swfObj.vjs_appendBuffer('throwDataSuperSecret');
   }
 
   /**
--- a/src/flash-source-buffer.js
+++ b/src/flash-source-buffer.js
@@ -25,6 +25,17 @@ const scheduleTick = function(func) {
 };
 
 /**
+ * Generates a random string of length 10
+ *
+ * @return {String} the randomly generated string
+ * @function generateRandomString
+ * @private
+ */
+const generateRandomString = function() {
+  return (Math.random().toString(36) + '00000000000000000').slice(2, 8);
+};
+
+/**
  * Round a number to a specified number of places much like
  * toFixed but return a number instead of a string representation.
  *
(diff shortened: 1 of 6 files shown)


src/flash-source-buffer.js
```diff
@@ -266,30 +274,27 @@ export default class FlashSourceBuffer extends videojs.EventTarget {
...
-    this.mediaSource_.swfObj.CallFunction(
-      '<invoke name="vjs_appendBuffer"' +
-      'returntype="javascript"><arguments><string>' +
-      b64str +
-      '</string></arguments></invoke>');
-
-    // schedule another append if necessary
-    if (this.bufferSize_ !== 0) {
+    /* eslint-disable camelcase */
+    window.vjs_flashEncodedData_ = function() {
```
imbcmdth: This function (and `vjs_flashEncodeHeader_` above) will collide if there are multiple players on the same page. We need some way to create unique names that we pass into Flash. I suggest the concating with the player id  and random value.

src/flash-source-buffer.js
```diff
@@ -77,18 +79,32 @@ export default class FlashSourceBuffer extends videojs.EventTarget {
...
+    window.vjs_flashEncodedHeader_ = function() {
+      delete window.vjs_flashEncodedHeader_;
+      return encodedHeader;
+    };
+    /* eslint-enable camelcase */
+
+    this.mediaSource_.swfObj.vjs_appendChunkReady('vjs_flashEncodedHeader_');
+
+    // TS to FLV transmuxer
+    this.transmuxer_ = work(transmuxWorker);
```
imbcmdth: Is `this.transmuxer_` terminated anywhere? Webworkers have to explicitly shut down or they live forever (or until the user navigates away from the page).

src/flash-source-buffer.js
```diff
@@ -266,30 +274,27 @@ export default class FlashSourceBuffer extends videojs.EventTarget {
...
-      this.updating = false;
-      this.trigger({ type: 'updateend' });
+      delete window.vjs_flashEncodedData_;
+      return b64str;
+    }.bind(this);
+    /* eslint-enable camelcase */
 
-    }
+    // bypass normal ExternalInterface calls and pass xml directly
+    // IE can be slow by default
```
imbcmdth: This comment doesn't describe what is happening anymore.

src/flash-source-buffer.js
mjneil: No it isn't. I see that for the html media source, the worker is terminated on the `sourceclose` event. Does flash emit this event as well and would it be the right time to terminate the worker for flash as well?

src/flash-source-buffer.js
```diff
@@ -23,6 +25,17 @@ const scheduleTick = function(func) {
...
 
 /**
+ * Generates a random string of length 10
+ *
+ * @return {String} the randomly generated string
+ * @function generateRandomString
+ * @private
+ */
+const generateRandomString = function() {
+  return (Math.random().toString(36) + '00000000000000000').slice(2, 8);
```
imbcmdth: Why add the zeros? Any value the string could take has a `1 / 2^52` chance of occurring - even the potential empty string. Also, why are there 17 zeros when you only take the first 10 characters that might be generated?

mjneil: This is a good point. The 0s help to guarantee that a shorter string won't be returned. But we don't really care about string length, just having a random string. 

(As for the 17 zeros, forgive me for I have copy pasted from stackoverflow)

imbcmdth:
LGTM

src/flash-transmuxer-worker.js
```diff
@@ -0,0 +1,122 @@
...
+  /**
+   * Recreate the transmuxer so that the next segment added via `push`
+   * start with a fresh transmuxer.
+   */
+  reset() {
+    this.init();
+  }
+
+  /**
+   * Forces the pipeline to finish processing the last segment and emit it's
```
gesinger: its

src/flash-transmuxer-worker.js
```diff
@@ -0,0 +1,122 @@
...
+   * results.
+   *
+   * @param {Object} data event data, not really used
```
gesinger: Probably don't need to mention it/include it in params.

src/flash-transmuxer-worker.js
```diff
@@ -0,0 +1,122 @@
...
+   */
+  flush(data) {
+    this.transmuxer.flush();
+  }
+}
+
+/**
+ * Our web wroker interface so that things can talk to mux.js
+ * that will be running in a web worker. the scope is passed to this by
```
gesinger: T

test/flash.test.js
```diff
@@ -66,46 +67,45 @@ const unfakeSTO = function() {
...
-      return output;
-    }, []);
-
-    datas.length = 0;
-    this.trigger('data', {
-      tags: {
-        videoTags: tags,
-        audioTags: []
+// Create a WebWorker-style message that signals the transmuxer is done
+const createDataMessage = function(datas, audioDatas, metadata, captions) {
```
gesinger: datas sounds funny to me. Maybe dataArray and audioDataArray? Or just data and audioData and it is implied it is an array?

test/flash.test.js
```diff
@@ -228,6 +225,8 @@ QUnit.test('creates FlashSourceBuffers for video/mp2t', function() {
 QUnit.test('waits for the next tick to append', function() {
   let sourceBuffer = this.mediaSource.addSourceBuffer('video/mp2t');
 
+  sourceBuffer.transmuxer_.postMessage = postMessage_;
```
gesinger: Should these replacements be moved into beforeEach, since it seems to be used in most places?

src/flash-source-buffer.js
```diff
@@ -77,18 +90,43 @@ export default class FlashSourceBuffer extends videojs.EventTarget {
...
+    // measure for pages with multiple players since these functions will be global
+    // instead of per instance.
+    this.flashEncodedHeaderName_ = 'vjs_flashEncodedHeader_' +
+                                   this.mediaSource_.player_.id() +
+                                   generateRandomString();
+    this.flashEncodedDataName_ = 'vjs_flashEncodedData_' +
+                                   this.mediaSource_.player_.id() +
+                                   generateRandomString();
+
+    /* eslint-disable camelcase */
```
mjneil: good call, it was spotting `window.vjs_flashEncodedData_` before the randomness change happened

src/flash-source-buffer.js
```diff
@@ -136,6 +174,10 @@ export default class FlashSourceBuffer extends videojs.EventTarget {
       removeCuesFromTrack(0, Infinity, this.metadataTrack_);
       removeCuesFromTrack(0, Infinity, this.inbandTextTrack_);
     });
+
+    this.mediaSource_.player_.tech_.hls.on('dispose', () => {
```
mjneil: That would mean adding a `dispose` method to the flash and virtual buffer, which is not a standard api on mediasource buffers

test/flash.test.js
mjneil: We can't because the sourceBuffer isn't created in the beforeEach since not every test uses the same MIME type

PR #128, opened 2017-02-04
Don't swallow append exceptions

dmlap:
Pass along exceptions that are triggered when appending data as a property on the synthetic player error.


imbcmdth:
LGTM

PR #131, opened 2017-02-10
Worker rename

balmasich:
Worker rename that allow use in webpack and webworkify-webpack-dropin


mjneil:
Thanks for the contribution @balmasich. We updated the name of the worker in the most recent version to bring in a web worker for flash transmuxing. Can you try using v4.4.0 and see if it solves your webpack issues, or if they still persist, could you update your PR with the needed changes to fix the issues.

balmasich:
I think now we can close this PR, because now in master branch is has renamed worker

PR #132, opened 2017-02-13
Revert worker for flash transmuxing

mjneil:
There were some issues with flash playback after #126 so this PR takes out the transmuxing in a web worker to be reviewed later and corrects the logic for trimming video tags to account for metadata tags added before key frames by the muxer

requires https://github.com/videojs/mux.js/pull/141

## Note
This should not be merged in until https://github.com/videojs/video-js-swf/pull/219 has been merged, released, and the swf url is updated in video.js

### DIFF
--- a/src/flash-source-buffer.js
+++ b/src/flash-source-buffer.js
@@ -7,8 +7,6 @@ import flv from 'mux.js/lib/flv';
 import removeCuesFromTrack from './remove-cues-from-track';
 import createTextTracksIfNecessary from './create-text-tracks-if-necessary';
 import {addTextTrackData} from './add-text-track-data';
-import transmuxWorker from './flash-transmuxer-worker';
-import work from 'webworkify';
 import FlashConstants from './flash-constants';
 
 /**
@@ -123,13 +121,8 @@ export default class FlashSourceBuffer extends videojs.EventTarget {
     this.mediaSource_.swfObj.vjs_appendChunkReady(this.flashEncodedHeaderName_);
 
     // TS to FLV transmuxer
-    this.transmuxer_ = work(transmuxWorker);
-    this.transmuxer_.postMessage({ action: 'init', options: {} });
-    this.transmuxer_.onmessage = (event) => {
-      if (event.data.action === 'data') {
-        this.receiveBuffer_(event.data.segment);
-      }
-    };
+    this.transmuxer_ = new flv.Transmuxer();
+    this.transmuxer_.on('data', this.receiveBuffer_.bind(this));
 
     this.one('updateend', () => {
       this.mediaSource_.tech_.trigger('loadedmetadata');
@@ -142,13 +135,15 @@ export default class FlashSourceBuffer extends videojs.EventTarget {
       set(val) {
         if (typeof val === 'number' && val >= 0) {
           this.timestampOffset_ = val;
+
+          this.transmuxer_ = new flv.Transmuxer();
+          this.transmuxer_.on('data', this.receiveBuffer_.bind(this));
+
           // We have to tell flash to expect a discontinuity
           this.mediaSource_.swfObj.vjs_discontinuity();
           // the media <-> PTS mapping must be re-established after
           // the discontinuity
           this.basePtsOffset_ = NaN;
-
-          this.transmuxer_.postMessage({ action: 'reset' });
         }
       }
     });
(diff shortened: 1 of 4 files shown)


src/flash-source-buffer.js
```diff
@@ -374,6 +371,19 @@ export default class FlashSourceBuffer extends videojs.EventTarget {
...
 
+    // gets the index of the key frame based on whether the tag is a metadata tag or not
+    const getKeyFrameIndex = (tag, index) => {
+      if (tag.metaDataTag) {
+        return index + 2;
+      }
+      return index;
+    };
+
+    // verifies the given index is valid index and points to a key frame
```
gesinger: extra "index"

src/flash-source-buffer.js
```diff
@@ -382,16 +392,29 @@ export default class FlashSourceBuffer extends videojs.EventTarget {
...
         while (nextIndex < videoTags.length) {
           let nextTag = videoTags[nextIndex];
 
           if (nextTag.pts >= targetPts) {
             break;
-          } else if (nextTag.keyFrame) {
+          } else if (nextTag.metaDataTag || nextTag.keyFrame) {
+            let nextKeyFrameIndex = getKeyFrameIndex(nextTag, nextIndex);
+
+            if (!verifyKeyFrameIndex(videoTags, nextKeyFrameIndex)) {
```
gesinger: Can we just reduce this condition and below to:

```js
foundNextKeyFrame = verifyKeyFrameIndex(videoTags, nextKeyFrameIndex);
break;
```

src/flash-source-buffer.js
```diff
@@ -382,17 +392,26 @@ export default class FlashSourceBuffer extends videojs.EventTarget {
 
       if (startTag.pts >= targetPts) {
         filteredVideoTags.push(startTag);
-      } else if (startTag.keyFrame) {
-        let nextIndex = startIndex + 1;
+      } else if (startTag.metaDataTag || startTag.keyFrame) {
+        let keyFrameIndex = getKeyFrameIndex(startTag, startIndex);
```
gesinger: Minor, but it seems a bit odd to me that we get the key frame index, but it's only the key frame index sometimes (from the call to verify below).

mjneil: verifying the index is mostly a sanity check as I think its possible (but I have not confirmed) that the two metadata tags could be followed by a non key frame. (I think if the segment does not start with a key frame. Here is the related mux.js code snippet

```
// Check if keyframe and the length of tags.
// This makes sure we write metadata on the first frame of a segment.
if (config && track && track.newMetadata &&
   (frame.keyFrame || tags.length === 0)) {
  // Push extra data on every IDR frame in case we did a stream change + seek
  tags.push(metaDataTag(config, frame.dts).finalize());
  tags.push(extraDataTag(track, frame.dts).finalize());
  track.newMetadata = false;
}
```

if `track.newMetadata === true && tags.length === 0 && frame.keyFrame === false`, the first tag being added would have the two metadata tags before it, but not be a key frame. By having that check, we wont treat a portion of tags as a GOP when they shouldn't be.

src/flash-source-buffer.js
```diff
@@ -422,26 +415,63 @@ export default class FlashSourceBuffer extends videojs.EventTarget {
...
       return;
     }
 
+    // If we are appending data that comes before our target pts, we want to tell
+    // the swf to adjust its notion of current time to account for the extra tags
+    // we are appending to complete the GOP that intersects with targetPts
+    if (tags[0].pts < targetPts && tech.seeking()) {
+      let currentTime = tech.currentTime();
+      let diff = (targetPts - tags[0].pts) / 1e3;
+      let adjustment = currentTime - diff;
```
gesinger: Minor: I'd call this adjustedTime, since `diff` seems more like the adjustment

src/flash-source-buffer.js
```diff
@@ -367,12 +364,18 @@ export default class FlashSourceBuffer extends videojs.EventTarget {
     targetPts *= 1e3;
     targetPts += this.basePtsOffset_;
 
-    // skip tags with a presentation time less than the seek target/end of buffer
-    for (let i = 0; i < audioTags.length; i++) {
-      if (audioTags[i].pts >= targetPts) {
-        filteredAudioTags.push(audioTags[i]);
+    // gets the index of the key frame based on whether the tag is a metadata tag or not
+    const getKeyFrameIndex = (tag, index) => {
```
imbcmdth: These functions should just be defined once at the top of the module. They don't appear to rely on closure variables and the overhead of recreating them each time `convertTagsToData_` is executed is wasteful - especially in IE where there is already a poorly performing VM.

src/flash-source-buffer.js
```diff
@@ -382,17 +385,26 @@ export default class FlashSourceBuffer extends videojs.EventTarget {
...
+
+        if (!verifyKeyFrameIndex(videoTags, keyFrameIndex)) {
+          startIndex++;
+          continue;
+        }
+
+        let nextIndex = keyFrameIndex + 1;
         let foundNextKeyFrame = false;
 
         while (nextIndex < videoTags.length) {
```
imbcmdth: This whole thing really needs some comments. What is the purpose of this inner loop? The outer loop?

mjneil:
Closing this PR because I separated it into two different PRs, one for the worker revert, and one for the other tag trimming bug fixes
https://github.com/videojs/videojs-contrib-media-sources/pull/133
https://github.com/videojs/videojs-contrib-media-sources/pull/134

PR #134, opened 2017-02-16
Fix flv tag trimming

mjneil:
corrects the logic for trimming video tags to account for metadata tags added before key frames by the muxer and

## Specific Enhancements
* uses `metaDataTag` flags on flv tags to make sure meta data for new GOPs are not dropped
* Notifies the swf to adjust it's internal notion of currentTime to account for extra data being appended when we are not trimming on a GOP boundary
* When the target trim timestamp intersects a GOP, we append the entire GOP which appends tags that come before the trim timestamp. Trimming audio tags did not take into account this extra data and could cause brief moments of silence when seeking. This is now taken into account when trimming the audio tags

## Note
* this is branched off of https://github.com/videojs/videojs-contrib-media-sources/pull/133 which should be merged in before this pr
* requires https://github.com/videojs/mux.js/pull/141
* This should not be merged in until https://github.com/videojs/video-js-swf/pull/219 has been merged, released, and the swf url is updated in video.js


mjneil:
Closing this PR because I separated it into two different PRs, one for the worker revert, and one for the other tag trimming bug fixes
https://github.com/videojs/videojs-contrib-media-sources/pull/133
https://github.com/videojs/videojs-contrib-media-sources/pull/134

PR #135, opened 2017-02-21
Revert "Revert flash transmuxing in a web worker (#133)"

mjneil:
Bringing the flash transmuxer in a web worker back round 2


forbesjo:
Looks good to me

PR #137, opened 2017-02-28
Fix flash tag trimming to not append duplicate tags

mjneil:
#136 introduced a change to track the last pts values of audio and video tags appended to the buffer to use as the target trim point for the next segment. This caused duplicate tags (usually the final GOP of the last segment appended) to be appended due to a `>=`. This change adds a fudge factor of `0.1` to the target pts when using the last stored pts value to avoid mistakenly appending a duplicate GOP

### DIFF
--- a/src/flash-source-buffer.js
+++ b/src/flash-source-buffer.js
@@ -377,7 +377,12 @@ export default class FlashSourceBuffer extends videojs.EventTarget {
       videoTargetPts *= 1e3;
       videoTargetPts += this.basePtsOffset_;
     } else {
-      videoTargetPts = this.videoBufferEnd_;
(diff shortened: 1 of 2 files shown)


src/flash-source-buffer.js
```diff
@@ -377,7 +377,7 @@ export default class FlashSourceBuffer extends videojs.EventTarget {
       videoTargetPts *= 1e3;
       videoTargetPts += this.basePtsOffset_;
     } else {
-      videoTargetPts = this.videoBufferEnd_;
+      videoTargetPts = this.videoBufferEnd_ + 0.1;
```
forbesjo: I feel like these need comments

forbesjo:
Looks good to me

PR #140, opened 2017-04-11
Support for multiple CEA608 tracks

squarebracket:
This change supports the work done in https://github.com/videojs/mux.js/pull/150. However, it should be backwards compatible with older mux.js versions.

Rather than `sourceBuffer.inbandTextTrack_` being a single object, it is now `sourceBuffer.inbandTextTracks_`, which is a map that stores `track => object`. When new captions come in, `caption.track` is inspected, and cues are added to the appropriate track.

### DIFF
--- a/src/cleanup-text-tracks.js
+++ b/src/cleanup-text-tracks.js
@@ -26,6 +26,15 @@ export const removeExistingTrack = function(player, kind, label) {
  * @private
  */
 export const cleanupTextTracks = function(player) {
-  removeExistingTrack(player, 'captions', 'cc1');
+  const tracks = player.remoteTextTracks() || [];
+
+  for (let i = tracks.length - 1; i >= 0; i--) {
+    const track = tracks[i];
+
+    if (track.kind === 'captions') {
+      player.removeRemoteTextTrack(track);
+    }
+  }
+
   removeExistingTrack(player, 'metadata', 'Timed Metadata');
 };
--- a/src/create-text-tracks-if-necessary.js
+++ b/src/create-text-tracks-if-necessary.js
@@ -16,14 +16,32 @@ const createTextTracksIfNecessary = function(sourceBuffer, mediaSource, segment)
 
   // create an in-band caption track if one is present in the segment
   if (segment.captions &&
-      segment.captions.length &&
-      !sourceBuffer.inbandTextTrack_) {
-    removeExistingTrack(player, 'captions', 'cc1');
-    sourceBuffer.inbandTextTrack_ = player.addRemoteTextTrack({
-      kind: 'captions',
-      label: 'cc1'
-    }, false).track;
-    player.tech_.trigger({type: 'usage', name: 'hls-608'});
+      segment.captions.length) {
+    if (!sourceBuffer.inbandTextTracks_) {
+      sourceBuffer.inbandTextTracks_ = {};
+    }
+
+    for (let trackId in segment.captionStreams) {
+      if (!sourceBuffer.inbandTextTracks_[trackId]) {
+        player.tech_.trigger({type: 'usage', name: 'hls-608'});
+        let track = player.textTracks().getTrackById(trackId);
+
+        if (track) {
+          // Resuse an existing track with a CC# id because this was
+          // very likely created by videojs-contrib-hls from information
+          // in the m3u8 for us to use
+          sourceBuffer.inbandTextTracks_[trackId] = track;
+        } else {
+          // Otherwise, create a track with the default `CC#` label and
+          // without a language
+          sourceBuffer.inbandTextTracks_[trackId] = player.addRemoteTextTrack({
+            kind: 'captions',
+            id: trackId,
+            label: trackId
+          }, false).track;
+        }
+      }
+    }
   }
 
   if (segment.metadata &&
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
   ],
   "dependencies": {
     "global": "^4.3.0",
-    "mux.js": "4.1.5",
+    "mux.js": "4.2.0",
     "video.js": "^5.17.0",
     "webworkify": "1.0.2"
   },
--- a/src/add-text-track-data.js
+++ b/src/add-text-track-data.js
@@ -64,7 +64,9 @@ const addTextTrackData = function(sourceHandler, captionArray, metadataArray) {
 
   if (captionArray) {
     captionArray.forEach(function(caption) {
-      this.inbandTextTrack_.addCue(
+      let track = caption.stream;
+
+      this.inbandTextTracks_[track].addCue(
         new Cue(
           caption.startTime + this.timestampOffset,
           caption.endTime + this.timestampOffset,
--- a/src/flash-source-buffer.js
+++ b/src/flash-source-buffer.js
@@ -179,7 +179,11 @@ export default class FlashSourceBuffer extends videojs.EventTarget {
     // of a buffered-range and everything else is reset on seek
     this.mediaSource_.player_.on('seeked', () => {
       removeCuesFromTrack(0, Infinity, this.metadataTrack_);
-      removeCuesFromTrack(0, Infinity, this.inbandTextTrack_);
+      if (this.inbandTextTracks_) {
+        for (let track in this.inbandTextTracks_) {
+          removeCuesFromTrack(0, Infinity, this.inbandTextTracks_[track]);
+        }
+      }
     });
 
     this.mediaSource_.player_.tech_.hls.on('dispose', () => {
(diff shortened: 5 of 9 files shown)


src/cleanup-text-tracks.js
```diff
@@ -26,6 +26,15 @@ export const removeExistingTrack = function(player, kind, label) {
  * @private
  */
 export const cleanupTextTracks = function(player) {
-  removeExistingTrack(player, 'captions', 'cc1');
+  const tracks = player.remoteTextTracks() || [];
+
+  for (let i = tracks.length - 1; i >= 0; i--) {
+    const track = tracks[i];
+
+    if (track.kind === 'captions' || track.kind === 'subtitles') {
```
mjneil: Why are tracks with `kind` subtitles` also removed here?

squarebracket: I'm not sure why I did this. I'm assuming subtitle tracks are cleaned elsewhere?

mjneil: They are supposed to be (this comment made me realize I've been using the `addRemoteTextTrack` api wrong all this time as they aren't actually getting properly cleaned). The subtitle tracks for webvtt are taken care of in `MasterPlaylistController` of contrib-hls.

src/create-text-tracks-if-necessary.js
```diff
@@ -16,14 +16,36 @@ const createTextTracksIfNecessary = function(sourceBuffer, mediaSource, segment)
...
-      kind: 'captions',
-      label: 'cc1'
-    }, false).track;
-    player.tech_.trigger({type: 'usage', name: 'hls-608'});
+      segment.captions.length) {
+    if (!sourceBuffer.inbandTextTracks_) {
+      sourceBuffer.inbandTextTracks_ = {};
+    }
+
+    segment.captions.forEach(function(caption) {
```
mjneil: This will have to loop through every caption of every segment loaded even after the text tracks have been created and added to the player. Though probably not a big deal, seems like a lot of wasted computation. Is there anyway we can attach to the transmuxers output the information of what streams the captions array uses so you can check whether theyve already been made without having to loop through the entire captions array again?

squarebracket: I'm not 100% sure I follow, so please do correct me if I've misunderstood. You're suggesting that the transmuxer keeps a local record of the tracks on which it's seen captions, and dispatches that information to media-sources? Like, say, when it dispatches a caption for a "new" track, it will set something like `caption.new = true`, which `media-sources` would then see and know it needs to create a new `remoteTextTrack` for that caption's `track` value?

That's certainly doable, and I agree that this loop on _every_ transmuxer push just feels wasteful. I'm not sure keeping state in the transmuxer is necessary, though. video.js historians might be more knowledgeable on this subject, but I'd hazard a guess that this code is only here anyway because the code in videojs/videojs-contrib-hls#1096 hasn't existed -- leaving this "track detection" code as the sole method for detecting and creating text tracks for CEA608 captions.

Perhaps some sort of logic could be included in videojs/videojs-contrib-hls#1096, something like, if closed captions are advertised in the playlist, then this code here would not be run, and we'd trust that all necessary caption tracks were declared in the playlist. If no text track exists, then we throw them into the void. And if captions aren't declared, but actually exist in the segments (I hope that is rare), then this loop would run.

That's potentially a breaking change from previous behaviour though, so please do let me know what you think.

src/cleanup-text-tracks.js
squarebracket: Yeah, that makes a lot more sense from what I know now about the architecture. I'll chock this one up to being a n00b mistake :)

src/create-text-tracks-if-necessary.js
mjneil: Something along those lines, just a matter of where and how to communicate that information. You're definitely right that this code exists because we haven't had a previous way of detecting which channels were in use until after we've transmuxed a segment. Since the CC tag information is optional in the manifest, having the detection and track creation here is still important. Trusting the manifest for optional information is usually a recipe for issues down the road.  I don't think the muxer would need to hold extra state as the information is already there in the captions. The transmuxer itself already needs to loop through the captions to adjust start and end times (https://github.com/videojs/mux.js/blob/master/lib/mp4/transmuxer.js#L973). This is also the last point that the transmuxer sees the captions before it sends it back to here. I think it would be reasonable to attach a bit of extra information to the `event` variable. Maybe something like `event.captionStreams` (or a more appropriate name) (Note: this would also have to be done in the `CoalesceStream` of the FLV' as well)
```
event.captionStreams = {
  CC1: false,
  CC2: false,
  CC3: false,
  CC4: false
};
```
and set the bool to true when you encounter it during that loop. Then instead of looping through the captions again here, you can just check `segment.captionStreams` and compare that to available tracks to determine if one needs to be made. I think we can avoid any breaking changes with this approach

squarebracket:
For some reason, I'm still getting a failure on a flash test. Printing out the event data in `createTextTracksIfNecessary`, I don't see `captionTracks`. Any idea why that might be? I've added the required stuff into the FLV transmuxer in mux.js....

mjneil:
I assume you mean one of the unit tests? The tests in this project pretty much entirely mock the output of the transmuxer. I suspect you need to edit this [function](https://github.com/videojs/videojs-contrib-media-sources/blob/master/test/flash.test.js#L71) to take some extra information and attach `captionTracks` to its return value.

squarebracket:
Ah, yeah, you're right. I saw mux.js being pulled in, but didn't see that you were overwriting certain functions :\

Ok, so I think everything is good now... although Firefox keeps dying in the Travis tests for some reason?

src/create-text-tracks-if-necessary.js
```diff
@@ -16,14 +16,34 @@ const createTextTracksIfNecessary = function(sourceBuffer, mediaSource, segment)
...
-      label: 'cc1'
-    }, false).track;
-    player.tech_.trigger({type: 'usage', name: 'hls-608'});
+      segment.captions.length) {
+    if (!sourceBuffer.inbandTextTracks_) {
+      sourceBuffer.inbandTextTracks_ = {};
+    }
+
+    for (let trackId in segment.captionTracks) {
+      let track = player.textTracks().getTrackById(trackId);
```
mjneil: Let's move this into the `if` block below. That way it will only go looking for the track on the player if the source buffer doesn't already have a `inbandTextTrack` stored for the `trackId`

mjneil:
I restarted the travis test. Looks like Firefox kept timing out. Usually re-running the test resolves that

src/create-text-tracks-if-necessary.js
squarebracket: Yup, great point, done.

squarebracket:
Firefox v55 was just released like a couple days ago, wasn't it? Maybe something is breaking there. I'm still running v54 and it's working for me.

mjneil:
Oh that could be related. I'm also on 54 locally, I'll try updating and see if that causes failure locally. If not I'll probably open in issue with TravisCI

mjneil:
FF55 locally still passing. submitted https://github.com/travis-ci/travis-ci/issues/8242 we'll see what they say

mjneil:
I just pushed to master a change to travis.yml to lock firefox at v54. Can you please rebase on top of master, it looks like restarting this travis build it is still picking up latest instead of 54

squarebracket:
Done.

src/create-text-tracks-if-necessary.js
```diff
@@ -16,14 +16,34 @@ const createTextTracksIfNecessary = function(sourceBuffer, mediaSource, segment)
...
+
+        if (track) {
+          // Resuse an existing track with a CC# id because this was
+          // very likely created by videojs-contrib-hls from information
+          // in the m3u8 for us to use
+          sourceBuffer.inbandTextTracks_[trackId] = track;
+        } else {
+          // Otherwise, create a track with the default `CC#` label and
+          // without a language
+          removeExistingTrack(player, 'captions', trackId);
```
mjneil: Now that we are letting contrib-hls create tracks for CC and this `else` block will only be entered if there is no text track on the player with `trackId`, I think this line is no longer needed as there won't be an existing track with `trackId` to remove.

squarebracket: Question: if I'm adding the CC tracks in contrib-hls, should I also be removing them in contrib-hls when the source changes?

mjneil: Good call. We shouldn't have to worry about removing them if we pass `false` as the second parameter to `addRemoteTextTrack`. I made a comment on your contrib-hls PR noting this.

squarebracket: Right, cool. Yeah in that case we should never have a situation where we have a `CC1` text track that has the wrong language/label.

Removing it caused [this test](https://github.com/squarebracket/videojs-contrib-media-sources/blob/d07b3a215ce5785b59ed4cf938b901ecb69bafd9/test/html.test.js#L1233) to fail, which makes sense. Think it's safe to just adapt the test, will deleting the `removeExistingTrack` have other possible side effects?

mjneil: Ya I think you should be safe just adapting the test. I don't think removing the call to `removeExistingTrack` here will have other side effects because of the way the `if` statements are laid out.

squarebracket:
Huh. Can you kick off that Travis test again?

mjneil:
Can you update package.json to use v4.2.0 of mux.js

squarebracket:
Done.

PR #142, opened 2017-04-27
Updateend events fire more than once on firefox

imbcmdth:
After an append, the SourceBuffer state (the `updating` property) can toggle to `false` long before the `updateend` events are triggered. The checks in `virtual-source-buffer` to "debounce" `updateend` events were inadequate to handle this scenario and would often trigger two `updateend` events instead of just one.

This should fix that scenario.

### DIFF
--- a/src/virtual-source-buffer.js
+++ b/src/virtual-source-buffer.js
@@ -469,14 +469,12 @@ export default class VirtualSourceBuffer extends videojs.EventTarget {
     if (this.videoBuffer_) {
       sortedSegments.video.segments.unshift(sortedSegments.video.initSegment);
       sortedSegments.video.bytes += sortedSegments.video.initSegment.byteLength;
-      this.pendingUpdateEnds_ += 1;
       this.concatAndAppendSegments_(sortedSegments.video, this.videoBuffer_);
       // TODO: are video tracks the only ones with text tracks?
       addTextTrackData(this, sortedSegments.captions, sortedSegments.metadata);
     }
 
     if (!this.audioDisabled_ && this.audioBuffer_) {
-      this.pendingUpdateEnds_ += 1;
       this.concatAndAppendSegments_(sortedSegments.audio, this.audioBuffer_);
     }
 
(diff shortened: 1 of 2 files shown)


src/virtual-source-buffer.js
```diff
@@ -457,11 +469,14 @@ export default class VirtualSourceBuffer extends videojs.EventTarget {
     if (this.videoBuffer_) {
       sortedSegments.video.segments.unshift(sortedSegments.video.initSegment);
       sortedSegments.video.bytes += sortedSegments.video.initSegment.byteLength;
+      this.pendingUpdateEnds_ += 1;
```
mjneil: This line (and removing corresponding audio one below) could be placed in the `try` block of `concatAndAppendSegments_` just before the call to `appendBuffer` on the native buffer

src/virtual-source-buffer.js
```diff
@@ -515,9 +555,11 @@ export default class VirtualSourceBuffer extends videojs.EventTarget {
    */
   abort() {
     if (this.videoBuffer_) {
+      this.videoBuffer_.updating = true;
```
mjneil: Why is this being added here (and for audio as well)?

According to the spec, when abort is called, if `buffer.updating === true` it should be set to `false` and an `updateend` event should be queued. If the real source buffer behind the wrapper does not have `updating === true` then an `updateend`event will never be fired and the wrapper will forever be `updating === true`

src/virtual-source-buffer.js
```diff
@@ -9,6 +9,31 @@ import work from 'webworkify';
...
+// We create a wrapper around the SourceBuffer so that we can manage the
+// state of the `updating` property manually. We have to do this because
+// Firefox changes `updating` to false long before triggering `updateend`
+// events and that was causing strange problems in videojs-contrib-hls
+const makeWrappedSourceBuffer = function(mediaSource, mimeType) {
+  const sourceBuffer = mediaSource.addSourceBuffer(mimeType);
+  const wrapper = Object.create(null);
+
+  wrapper.updating = false;
+  wrapper.realBuffer_ = sourceBuffer;
```
mjneil: My guess is for debugging purposes, it might be nice to be able to grab a reference to the actual buffer object

gesinger:
re: my question on aborting, @mjneil answered it by pointing to https://www.w3.org/TR/media-source/#dom-sourcebuffer-abort 👍

PR #145, opened 2017-05-24
chore: update to babel 6

gkatsev:


gkatsev:
This will just happen as part of the rollup PR #147.

PR #147, opened 2017-06-14
WIP: Rollup

gkatsev:


gkatsev:
Unfortunately, getting rid of the duplication that happens with rolluping up each worker is causing a too-large size increase to be able to proceed with this right now. Closing for now until we can get back it.

PR #151, opened 2017-07-05
fix: Don't process segments if the mediaSource is closed

forbesjo:
In certain situations we might try to append segment data when the mediaSource is closed.

Todo: add tests

### DIFF
--- a/src/virtual-source-buffer.js
+++ b/src/virtual-source-buffer.js
@@ -263,6 +263,12 @@ export default class VirtualSourceBuffer extends videojs.EventTarget {
    * @param {Event} event the done event from the transmuxer
    */
   done_(event) {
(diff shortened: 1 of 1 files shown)


src/virtual-source-buffer.js
```diff
@@ -263,6 +263,12 @@ export default class VirtualSourceBuffer extends videojs.EventTarget {
    * @param {Event} event the done event from the transmuxer
    */
   done_(event) {
+    // Don't process and append data if the mediaSource is closed
+    if (this.mediaSource_.readyState === 'closed') {
```
gesinger: How did we get to the point where we're trying to append data even though the media source is closed?

forbesjo: Going to investigate into this further.

PR #152, opened 2017-07-10
Fix disabled audio for a single audio only source buffer

mjneil:
When loading an audio only playlist, we would create a single source buffer that had only audio codecs. When changes to the audio tracks happen, the `HtmlMediaSource` would update which source buffer has ownership of audio. When the `main` audio track is enabled, we used the assumption that audio is muxed into the main playlist, and alternate audio is not being used, so audio would be enabled in the "combined" source buffer, and disabled in the "audio only" source buffer. However, in the case of audio only playlist, the "combined" source buffer would only have an audio codec, which was incorrectly being marked as the "audio only" source buffer, so audio would be disabled and prevent data from ever being appended.

This problem was hidden by a race condition between contrib-hls `MasterPlaylistController` completing the first media playlist request and the `MediaSource` triggering the `sourceopen` event. If the `sourceopen` event happened second, the audio track change event listeners would be setup later, avoiding this bug.

This change checks to see if there is only one source buffer, and if so make sure it is always active and enables audio based on the source buffer's audio codec information.

Addresses:
https://github.com/videojs/videojs-contrib-hls/issues/1186
https://github.com/videojs/videojs-contrib-hls/issues/1179

Audio only in Flash also has some issues when we detect that the type is `audio/mp2t`. This address that by allowing audio only `FlashSourceBuffer`s to be created.

### DIFF
--- a/src/flash-media-source.js
+++ b/src/flash-media-source.js
@@ -72,7 +72,7 @@ export default class FlashMediaSource extends videojs.EventTarget {
     let sourceBuffer;
 
     // if this is an FLV type, we'll push data to flash
-    if (parsedType.type === 'video/mp2t') {
+    if (parsedType.type === 'video/mp2t' || parsedType.type === 'audio/mp2t') {
       // Flash source buffers
       sourceBuffer = new FlashSourceBuffer(this);
     } else {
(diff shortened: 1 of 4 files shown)


forbesjo:
Looks good to me

PR #154, opened 2017-08-11
lock travis builds to use firefox v54

mjneil:
https://github.com/travis-ci/travis-ci/issues/8242

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ env:
   - VJS=5
   - VJS=6
 addons:
-  firefox: latest
+  firefox: "54.0"
   apt:
     sources:
     - google-chrome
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -22,7 +22,7 @@ env:
   - VJS=5
   - VJS=6
 addons:
-  firefox: latest
+  firefox: "54.0"
```
gesinger: We should add a comment with the Travis issue and a TODO, because this is fragile for users, since most Firefox users will be on a more modern version (meaning out tests should use more modern versions).

mjneil: This is definitely meant to be temporary until the issue is resolved, I'll add a TODO

PR #155, opened 2017-08-14
update travis to use latest firefox

mjneil:


test/karma.conf.js
```diff
@@ -51,6 +51,8 @@ module.exports = function(config) {
       }
     },
     detectBrowsers: detectBrowsers,
+    // timeout after 2 minutes of no activity
```
gesinger: Why was 2 minutes chosen?

mjneil: Experimenting increase of timeout per https://github.com/travis-ci/travis-ci/issues/8242#issuecomment-322199112

gesinger: Ah, OK, didn't realize it was for experimentation.

PR #156, opened 2017-08-14
fix duplicate caption processing

squarebracket:
@mjneil I think this is how I did it, but I don't recall for sure tbh.

### DIFF
--- a/src/flash-source-buffer.js
+++ b/src/flash-source-buffer.js
@@ -182,9 +182,16 @@ export default class FlashSourceBuffer extends videojs.EventTarget {
       removeCuesFromTrack(0, Infinity, this.inbandTextTrack_);
     });
 
+    this.mediaSource_.player_.tech_.on('hls-reset', this.resetHls);
+
     this.mediaSource_.player_.tech_.hls.on('dispose', () => {
       this.transmuxer_.terminate();
+      this.mediaSource_.player_.tech_.off('hls-reset', this.resetHls);
     });
+  }
+
+  resetHls() {
+    this.transmuxer_.postMessage({action: 'resetCaptions'});
   }
 
   /**
--- a/src/flash-source-buffer.js
+++ b/src/flash-source-buffer.js
@@ -182,16 +182,14 @@ export default class FlashSourceBuffer extends videojs.EventTarget {
       removeCuesFromTrack(0, Infinity, this.inbandTextTrack_);
     });
 
-    this.mediaSource_.player_.tech_.on('hls-reset', this.resetHls);
+    let onHlsReset = this.onHlsReset_.bind(this);
+
+    this.mediaSource_.player_.tech_.on('hls-reset', onHlsReset);
 
     this.mediaSource_.player_.tech_.hls.on('dispose', () => {
       this.transmuxer_.terminate();
-      this.mediaSource_.player_.tech_.off('hls-reset', this.resetHls);
+      this.mediaSource_.player_.tech_.off('hls-reset', onHlsReset);
     });
-  }
-
-  resetHls() {
-    this.transmuxer_.postMessage({action: 'resetCaptions'});
   }
 
   /**
@@ -545,5 +543,9 @@ export default class FlashSourceBuffer extends videojs.EventTarget {
     }
 
     return tags;
+  }
+
+  onHlsReset_() {
+    this.transmuxer_.postMessage({action: 'resetCaptions'});
   }
 }
--- a/src/html-media-source.js
+++ b/src/html-media-source.js
@@ -159,6 +159,14 @@ export default class HtmlMediaSource extends videojs.EventTarget {
       });
     };
 
+    this.onResetHls_ = () => {
+      this.sourceBuffers.forEach((sourceBuffer) => {
+        if (sourceBuffer.transmuxer_) {
+          sourceBuffer.transmuxer_.postMessage({action: 'resetCaptions'});
+        }
+      });
+    };
+
     // Re-emit MediaSource events on the polyfill
     [
       'sourceopen',
@@ -179,6 +187,8 @@ export default class HtmlMediaSource extends videojs.EventTarget {
       }
 
       this.player_ = videojs(video.parentNode);
+
+      this.player_.tech_.on('hls-reset', this.onResetHls_);
 
       if (this.player_.audioTracks && this.player_.audioTracks()) {
         this.player_.audioTracks().on('change', this.updateActiveSourceBuffers_);
@@ -230,6 +240,7 @@ export default class HtmlMediaSource extends videojs.EventTarget {
       // event handlers left to unbind anyway
       if (this.player_.el_) {
         this.player_.off('mediachange', this.onPlayerMediachange_);
+        this.player_.tech_.off('hls-reset', this.onResetHls_);
       }
     });
   }
--- a/src/html-media-source.js
+++ b/src/html-media-source.js
@@ -159,7 +159,7 @@ export default class HtmlMediaSource extends videojs.EventTarget {
       });
     };
 
-    this.onResetHls_ = () => {
+    this.onHlsReset_ = () => {
       this.sourceBuffers.forEach((sourceBuffer) => {
         if (sourceBuffer.transmuxer_) {
           sourceBuffer.transmuxer_.postMessage({action: 'resetCaptions'});
@@ -188,7 +188,7 @@ export default class HtmlMediaSource extends videojs.EventTarget {
 
       this.player_ = videojs(video.parentNode);
 
-      this.player_.tech_.on('hls-reset', this.onResetHls_);
+      this.player_.tech_.on('hls-reset', this.onHlsReset_);
 
       if (this.player_.audioTracks && this.player_.audioTracks()) {
         this.player_.audioTracks().on('change', this.updateActiveSourceBuffers_);
@@ -240,7 +240,7 @@ export default class HtmlMediaSource extends videojs.EventTarget {
       // event handlers left to unbind anyway
       if (this.player_.el_) {
         this.player_.off('mediachange', this.onPlayerMediachange_);
-        this.player_.tech_.off('hls-reset', this.onResetHls_);
+        this.player_.tech_.off('hls-reset', this.onHlsReset_);
       }
     });
   }
--- a/src/flash-transmuxer-worker.js
+++ b/src/flash-transmuxer-worker.js
@@ -76,6 +76,11 @@ class MessageHandlers {
   flush() {
     this.transmuxer.flush();
   }
+
+  resetCaptions() {
+    this.transmuxer.resetCaptions();
+  }
+
 }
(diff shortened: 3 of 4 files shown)


mjneil:
I would setup the `seeked` handler within the `HtmlMediaSource` `sourceopen` handler (https://github.com/videojs/videojs-contrib-media-sources/blob/master/src/html-media-source.js#L173) as that is where `HtmlMediaSource.player_` is attached which may or may not be available at time of `VirtualSourceBuffer` construction. You could define a function similar to `this.onPlayerMediaChange_` that loops through the sourcebuffers and calling the transmuxer api if it's attached to the source buffer and `off` the handler in the corresponding `sourceclose` handler. You'll have to do similar implementations for the `FlashMediaSource` and `FlashTransmuxWorker`.

squarebracket:
How's this?

I added the flash stuff, but I'm not sure what I did was appropriate. There's no handler for `sourceclose`, so I added one. But maybe I should have stuck it in `endOfStream`?

squarebracket:
Ah, my Chrome keeps "aw, snap!"-ing, so I didn't know the tests were failing. Will look into it.

squarebracket:
Ok, nevermind, apparently all the tests are fine.... :thinking:

mjneil:
Here's where I tell you I steered you in the wrong direction because flash is special haha. Since flash is not using real MSE and instead the swf is acting like a media source, I'm not entirely sure if it is actually firing `sourceopen` and `sourceclose` events as we expect.  For this reason I think at least for the flash side, we can put the code in the `FlashSourceBuffer`. Specifcally, there is already a listener for `seeked` here https://github.com/videojs/videojs-contrib-media-sources/blob/master/src/flash-source-buffer.js#L180 that you can add your call to the transmuxer seeked api.
Since this `seeked` handler is already not being unregistered, you probably don't have to worry about that

src/html-media-source.js
```diff
@@ -187,6 +199,7 @@ export default class HtmlMediaSource extends videojs.EventTarget {
       }
 
       this.player_.on('mediachange', this.onPlayerMediachange_);
+      this.player_.on('seeking', this.onPlayerSeeking_);
```
mjneil: Why `seeking` instead of `seeked`?

src/html-media-source.js
```diff
@@ -159,6 +159,18 @@ export default class HtmlMediaSource extends videojs.EventTarget {
       });
     };
 
+    this.onPlayerSeeking_ = (event) => {
+      this.sourceBuffers.forEach((sourceBuffer) => {
+        if (sourceBuffer.transmuxer_) {
+          let range = this.player_.buffered();
+          let current = this.player_.currentTime();
+          if (current >= range.end(0) || current <= range.start(0)) {
```
mjneil: You should check `range.length` before attempting to call `start` or `end` because if the buffer is empty you'll get an error

src/html-media-source.js
squarebracket: I'm not sure if it matters, but the captions should all be reset before it starts ingesting the video. Don't know if the `seeked` event will fire and complete before data is `push`ed through the transmuxer.

src/html-media-source.js
squarebracket: Added `range.length && ` to the if statement.

src/html-media-source.js
mjneil: Hmmm better safe than sorry I suppose

src/html-media-source.js
```diff
@@ -159,6 +159,19 @@ export default class HtmlMediaSource extends videojs.EventTarget {
       });
     };
 
+    this.onPlayerSeeking_ = (event) => {
+      this.sourceBuffers.forEach((sourceBuffer) => {
+        if (sourceBuffer.transmuxer_) {
+          let range = this.player_.buffered();
+          let current = this.player_.currentTime();
+
+          if (current >= range.end(0) || current <= range.start(0)) {
```
mjneil: Since this won't change between buffers, we should do this check first and only do the `forEach` if it passes. 

You'll also want to check `range.length` here.

I think this comparison will also run into problems if there are gaps in the buffer. (history for context) In the past, contrib-hls had pretty complicated logic surrounding segment loading, and part of that complication was gaps and multiple time ranges after seeks. We've since updated that so contrib-hls appends segments in a sequential order and clears the entire buffer when seeking outside the buffered range. What this gave us was the ability for contrib-hls to have the assumption that there is only ever 1 time range of buffered content, and any gaps within that range are inherit to the source and not because we seeked or skipped a segment. However, `buffered()` would still reflect those gaps. For example buffered could look like `[0 -> 10, 11->20]` after appending the first two segments. Seeking to 13 would not reset anything from contrib-hls point of view, but `range.end(0)` would be `10` so this would cause the captions to be reset.

Specifically, contrib-hls will do its seek reset if [this](https://github.com/videojs/videojs-contrib-hls/blob/master/src/master-playlist-controller.js#L1360) `buffered.length === 0`. You'll notice the [findRange function](https://github.com/videojs/videojs-contrib-hls/blob/master/src/ranges.js#L46) it calls also uses a fudge factor to account for rounding of the time range values, so you should probably do something similar here as well

src/flash-source-buffer.js
```diff
@@ -178,6 +178,7 @@ export default class FlashSourceBuffer extends videojs.EventTarget {
     // On a seek we remove all text track data since flash has no concept
     // of a buffered-range and everything else is reset on seek
     this.mediaSource_.player_.on('seeked', () => {
+      this.resetCaptionsIfAppropriate();
```
mjneil: Just realized that things are reset on every seek regardless of in buffer or not when Flash is being used, so you should just be able to `postMessage` here without doing any checks

src/html-media-source.js
squarebracket: Hmm. Think maybe contrib-hls should dispatch the actual message, then? I mean, all this other infrastructure for handling the exact same case is there...

squarebracket: Or would that break things for native HLS?

mjneil: Having contrib-hls handle the message would simplify this alot. I agree we should come up with a solution that can take advantage of the code and information contrib-hls has.

Historically we've tried to keep this project as a shim for MSE so that contrib-hls would not need knowledge of whether it was working with this project or native media source. (Even though contrib-hls is always using our shim). Eventually we would like to do away with this project and pull functionality into contrib-hls so we don't have to deal with these constraints that we keep running into. However, re-examining how the two projects interact, we are already [breaking that principle by calling ](https://github.com/videojs/videojs-contrib-hls/blob/master/src/master-playlist-controller.js#L445)`addSeekableRange_` from contrib-hls. We also communicate `mediachange` between these projects through an [event on the tech](https://github.com/videojs/videojs-contrib-hls/blob/master/src/master-playlist-controller.js#L505) that [contrib-media-sources listens to](https://github.com/videojs/videojs-contrib-media-sources/blob/master/src/html-media-source.js#L189). I don't think we should add new apis for contrib-hls to call (even though we are doing it once). I think the event solution is reasonable and keeps with the spirit of being a native shim. If contrib-hls was working with a native mediasource, triggering an event on the tech would not break anything as there would just be nothing listening to it. 

So instead of doing the reset on seek events, lets listen for a custom event on the player, and do the reset when that event is triggered. Then all that would be needed is for contrib-hls to trigger the event at the right moment

mjneil: I think this is all that would need to be done from contrib-hls https://github.com/mjneil/videojs-contrib-hls/commit/6cfdb6b32950a88908954230d47d6d554484b5a2

squarebracket: I like that approach!

squarebracket: Is there a "global" place in media-sources where I could receive that event, or will I have to set up the listener for HTML and Flash separately?

mjneil: No I don't think so. You'd have to do them separately like it is now

squarebracket: Ok I pushed to squarebracket/videojs-contrib-media-sources@e21be4a75ff52bf976ebd3ddc26687aec5637dca and added your commit to contrib-hls, seems to work.

squarebracket: nevermind, must be some caching thing wrong with my environment...

squarebracket: Derp, I'm a dummy, was listening to `resetEverything` instead of `hls-reset-everything`. Working now.

mjneil: left a couple comments on that commit otherwise 👍

squarebracket:
Alright, changed things to the event-based method.

mjneil:
can we update the `hls-reset-everything` event to be just `hls-reset` (and have contrib-hls changed to fire `hls-reset` as well)

squarebracket:
Done. I renamed the callbacks as well.

src/flash-source-buffer.js
```diff
@@ -182,11 +182,18 @@ export default class FlashSourceBuffer extends videojs.EventTarget {
       removeCuesFromTrack(0, Infinity, this.inbandTextTrack_);
     });
 
+    this.mediaSource_.player_.tech_.on('hls-reset', this.resetHls);
```
mjneil: For some reason when `this.resetHls` is called after hearing this event, `this` within `resetHls` is not being bound correctly.

if you define the function above instead of using the es6 class syntax it works
```
this.resetHls = () => {
  this.transmuxer_.postMessage({action: 'resetCaptions'});
};

this.mediaSource_.player.tech_.on('hls-reset', .......
```

squarebracket: Looks like `bind` is being used in a couple other places there, maybe it's best to do that to keep the style consistent.

mjneil: bind would work too. You'll just have to make sure to put it in a local var and use that for both the `on` and `off` otherwise the references passed to `on` and `off` will be to different functions
```
let onHlsReset = this.resetHls.bind(this);

...tech_.on('hls-reset', onHlsReset);
...tech_.off('hls-reset', onHlsReset);
```

src/flash-source-buffer.js
```diff
@@ -182,11 +182,20 @@ export default class FlashSourceBuffer extends videojs.EventTarget {
...
+
+    this.mediaSource_.player_.tech_.on('hls-reset', onHlsReset);
+
     this.mediaSource_.player_.tech_.hls.on('dispose', () => {
       this.transmuxer_.terminate();
+      this.mediaSource_.player_.tech_.off('hls-reset', onHlsReset);
     });
   }
 
+  resetHls() {
```
mjneil: can we update this to `onHlsReset_` for naming consistency

src/html-media-source.js
```diff
@@ -159,6 +159,14 @@ export default class HtmlMediaSource extends videojs.EventTarget {
       });
     };
 
+    this.onResetHls_ = () => {
```
mjneil: can we also update this to `onHlsReset_` as well

PR #160, opened 2017-09-20
turn off firefox for travisci

mjneil:
https://github.com/travis-ci/travis-ci/issues/8242

### DIFF
--- a/test/karma.conf.js
+++ b/test/karma.conf.js
@@ -13,12 +13,11 @@ module.exports = function(config) {
     }
   };
 
-  // On Travis CI, we can only run in Firefox.
-  // TODO: This should be using firefox. It is currently turned off because
+  // TODO: This should include firefox. It is currently turned off because
   //       of https://github.com/travis-ci/travis-ci/issues/8242 When this issue is
   //       resolved, this should be updated back to latest
   if (process.env.TRAVIS) {
-    config.browsers = ['travisChrome'];
+    config.browsers = ['ChromeHeadless'];
   }
 
   // If no browsers are specified, we enable `karma-detect-browsers`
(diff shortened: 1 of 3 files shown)


test/karma.conf.js
```diff
@@ -15,7 +15,7 @@ module.exports = function(config) {
 
   // On Travis CI, we can only run in Firefox.
```
gesinger: This comment is a bit out of date. Would be worth mentioning why firefox isn't included.

forbesjo:
You can use a simpler Chrome set up now https://github.com/forbesjo/mpd-parser/blob/master/.travis.yml#L9
And maybe headless Chrome https://github.com/forbesjo/mpd-parser/blob/master/test/karma.conf.js

test/karma.conf.js
```diff
@@ -14,8 +14,11 @@ module.exports = function(config) {
   };
 
   // On Travis CI, we can only run in Firefox.
+  // TODO: This should be using firefox. It is currently turned off because
+  //       of https://github.com/travis-ci/travis-ci/issues/8242 When this issue is
+  //       resolved, this should be updated back to latest
```
gesinger: This should probably just say to include 'Firefox'

test/karma.conf.js
gesinger: Probably best to remove this comment.

test/karma.conf.js
```diff
@@ -13,9 +13,11 @@ module.exports = function(config) {
     }
   };
 
-  // On Travis CI, we can only run in Firefox.
+  // TODO: This should include firefox. It is currently turned off because
+  //       of https://github.com/travis-ci/travis-ci/issues/8242 When this issue is
+  //       resolved, this should be updated back to latest
```
gesinger: minor: but in this scenario it should just say `this should be updated to include firefox` instead of the mention of latest

gesinger: Or just remove everything after the issue link

forbesjo:
Bump "karma-chrome-launcher" to "^2.2.0" to get headless chrome support

Issue #167, opened 2017-12-14
Incorrect old video.js dependency version


eigood:
In a project I've started, I installed video.js, which pulled in the latest version, 6.x.  I then added videojs-contrib-hls, which depends on video.js 5.x || 6.x.  This then depends on videojs-contrib-media-sources, which depends on video-js 5.x.  So now I have 2 copies of video.js in my webpack built bundle.

If you could update the video.js dep version to be done like videojs-contrib-hls, that would help tremendously, and reduce the size built bundle.

gkatsev:
@eigood care to submit a PR that adds `|| 6.x` to the field in package.json?

akofman:
Hello, 
Maybe I missed something but is it actually necessary to keep this dependency ? I mean, why don't you keep it as a devDependency and indicate it as external to Browserify ?

gkatsev:
Because I've found that the peerDep system that previously was in npm to be unreliable at best and incredibly frustrating and broken at worst. Having it as a dependency isn't bad if the semver ranges inclusive enough as most bundlers, at least with the addition of `npm dedupe` (which happens automatically in newer npm etc) should take care of you not having multiple versions of video.js.

eigood:
Sure, hang on.

eigood:
Ping?

gkatsev:
Oops, totally forgot. Thanks for your patience, we'll take a look today.

PR #169, opened 2017-12-21
Update video.js dependency to allow for v6 to be installed.

eigood:
This is based on the same dep pattern that videojs-contrib-hls uses.
With this update, when external projects decide to use video.js v6,
there won't end up being 2 copies bundled into the output.

Closes: videojs/videojs-contrib-media-sources#167

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,7 +78,7 @@
   "dependencies": {
     "global": "^4.3.0",
     "mux.js": "4.3.2",
-    "video.js": "^5.17.0",
+    "video.js": "^5.17.0 || ^6.2.0",
     "webworkify": "1.0.2"
   },
   "devDependencies": {


eigood:
I'm attempting to switch my local project to this, but am having issues with webpack, probably because the built version isn't there, and when I attempt to alias imports to the src version, it's having issues.  I'm still looking into this.

eigood:
Since dist/. and es5/ are excluded via .gitignore(rightly so), and the method of publishing this isn't documented(via CONTRIBUTING.md), I don't have a way to test this locally.

I tried setting aliases in my webpack.config.js, but then ran into issues both in -hls and -media-sources, with the way each does their exports, as they aren't compatible with es6.

eigood:
npm pack, then npm install the tarball, and this works, reducing my player bundle(split, tree shook, uglify, gzip) from 497.1 to 453.1.

PR #171, opened 2018-03-01
Add text tracks for both video tracks and audio-only tracks

jamesjieye:
This is related to an issue I experienced that is the same as https://github.com/videojs/videojs-contrib-hls/issues/1255.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "videojs-contrib-media-sources",
-  "version": "4.6.2",
+  "version": "4.6.3",
   "description": "A Media Source Extensions plugin for video.js",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "videojs-contrib-media-sources",
-  "version": "4.6.2",
+  "version": "4.6.3",
```
gesinger: Usually we change the version as part of a versioning process (outside of the PR)

jamesjieye: I see. Let me fix this and create another PR.

jamesjieye:
Close this in favor of #172.

PR #174, opened 2018-03-06
update travis config

squarebracket:


test/karma.conf.js
```diff
@@ -46,13 +38,27 @@ module.exports = function(config) {
...
+      }
+    },
     detectBrowsers: detectBrowsers,
     reporters: ['dots'],
     port: 9876,
     colors: true,
     autoWatch: false,
     singleRun: true,
-    concurrency: Infinity,
+    concurrency: 1,
```
mjneil: Why were these values for `concurrency`, `browserNoActivityTimeout` and `browserDisconnectTimeout` added here when in https://github.com/videojs/videojs-contrib-dash/pull/229/files#diff-53913120a28c73adf6d34da9be8f4847L38 these values were removed?

squarebracket: Only because [they were added for VHS](https://github.com/videojs/http-streaming/blob/master/test/karma.conf.js).

They may not be needed here, but contrib-hls (and I assume VHS) tends to time out during tests. Do you want me to remove them?

squarebracket: (I believe @forbesjo took those values from video.js' config)

mjneil: nah leaving them is fine, was just wondering since it seemed to be the opposite of what the other pr did

PR #178, opened 2018-04-25
Graceful handling of QuotaExceededError

squarebracket:
This PR adds the media-sources-specific bits required for graceful handling of `QuotaExceededError`s.

Not too much has changed.

1. `removeCuesFromTrack` now only removes cues that are wholly within the removal region

2. The wrapped source buffer inside `virtual-source-buffer.js` now takes care of setting its `updating` property, so that it can be passed along to contrib-hls and treated like a normal source buffer.

3. Rather than throw an error, it now simply triggers a `bufferMaxed` error on itself, with the buffer that caused the error as the event's `target`.

4. Added a `currGOPStart` function to get the last GOP start from the GOP buffer, so that it can be included in the `bufferMaxed` event data.

Informing contrib-hls of all GOP start times would probably help with extreme circumstances, but I opted to leave that bit out for now.


src/remove-cues-from-track.js
```diff
@@ -27,8 +27,8 @@ const removeCuesFromTrack = function(start, end, track) {
   while (i--) {
     cue = track.cues[i];
 
-    // Remove any overlapping cue
-    if (cue.startTime <= end && cue.endTime >= start) {
+    // Remove any cue wholly within the removal region
+    if (cue.startTime >= start && cue.endTime <= end) {
```
ldayananda, line 31: Are there any possible issues with leaving out the overlapping cues from this condition?

src/virtual-source-buffer.js
```diff
@@ -13,26 +13,43 @@ import {isAudioCodec, isVideoCodec} from './codec-utils';
...
+            sourceBuffer[key](...params);
+          } catch (error) {
+            wrapper.updating = false;
+            throw error;
+          }
+        };
+      } else if (key === 'abort') {
+        wrapper[key] = (...params) => {
+          wrapper.updating = false;
+          sourceBuffer[key](...params);
```
ldayananda, line 38: should we be handling errors here too?

src/virtual-source-buffer.js
```diff
@@ -66,6 +83,42 @@ export const gopsSafeToAlignWith = (buffer, player, mapping) => {
...
+
+  let i;
+
+  for (i = 0; i < buffer.length; i++) {
+    if (buffer[i].pts > currentTimePts) {
+      break;
+    }
+  }
+
+  if (i === 0 || buffer[i - 1].pts > currentTimePts) {
```
ldayananda: Could use a comment here. What is `buffer[i - 1].pts > currentTimePts` checking for?

src/virtual-source-buffer.js
```diff
@@ -66,6 +83,42 @@ export const gopsSafeToAlignWith = (buffer, player, mapping) => {
...
+ * Returns the start time of the current GOP
+ *
+ * @param {Array} buffer
+ *        The current buffer of gop information
+ * @param {Player} player
+ *        The player instance
+ * @param {Double} mapping
+ *        Offset to map display time to stream presentation time
+ * @return {Double}
+ *         Time of current GOP
```
ldayananda: *Display* time of current GOP?

src/virtual-source-buffer.js
```diff
@@ -66,6 +83,42 @@ export const gopsSafeToAlignWith = (buffer, player, mapping) => {
...
+ */
+export const currentGOPStart = (buffer, player, mapping) => {
+  if (!player || !buffer.length) {
+    return null;
+  }
+
+  // pts value for current time
+  const currentTimePts = Math.ceil((player.currentTime() - mapping) * 90000);
```
ldayananda: Why not just take `currentTime` as an argument instead of a player instance? I feel like that's less likely to break

src/virtual-source-buffer.js
```diff
@@ -718,17 +769,18 @@ export default class VirtualSourceBuffer extends videojs.EventTarget {
...
-            type: 'APPEND_BUFFER_ERR',
-            message: error.message,
-            originalError: error
-          });
-        }
+        let currGOPStart = currentGOPStart(this.gopBuffer_,
+                                           this.mediaSource_.player_,
+                                           this.timeMapping_);
+
+        this.trigger({
```
ldayananda: Are there any other errors that might be triggered by `appendBuffer` that would be swallowed by this change? Some errors we want to trigger on the player, especially if they are "fatal" playback errors.

ldayananda: Seems like we should only trigger `bufferMaxed` if we see the `QuotaExceededError`?

test/html.test.js
```diff
@@ -386,11 +388,16 @@ QUnit.test('appendBuffer error triggers on the player', function() {
...
   sourceBuffer.videoBuffer_.appendBuffer = () => {
+    QUnit.equal(sourceBuffer.videoBuffer_.updating, true,
+      'updating is true before error');
     throw new Error();
   };
 
-  this.player.on('error', () => error = true);
-
+  sourceBuffer.on('bufferMaxed', (event) => {
+    QUnit.ok(true, 'event called');
```
ldayananda: technically the next assert tests this as well

test/html.test.js
```diff
@@ -1602,6 +1607,47 @@ QUnit.test('gopsSafeToAlignWith returns correct list', function() {
...
+  actual = currentGOPStart(buffer, player, mapping);
+  QUnit.deepEqual(actual, expected, 'null when entire buffer is ahead of currentTime');
+
+  currentTime = 1.5;
+  actual = currentGOPStart(buffer, player, mapping);
+  QUnit.deepEqual(actual, 1, 'uses previous GOP when between GOP starts');
+
+  currentTime = 2.2;
+  actual = currentGOPStart(buffer, player, mapping);
+  QUnit.deepEqual(actual, 2.200011111111111, 'uses currentTime when currentTime === GOP start');
```
ldayananda: this assertion threw me off as I didn't notice `pts` was using the ceiling of the time input at first

test/html.test.js
```diff
@@ -1602,6 +1607,47 @@ QUnit.test('gopsSafeToAlignWith returns correct list', function() {
     'empty array when no gops in buffer come after currentTime');
 });
 
+QUnit.test('currentGOPStart returns time of most recent GOP', function(assert) {
+  const pts = (time) => Math.ceil(time * 90000);
```
ldayananda: To avoid confusion, let's not use the ceiling of the pts values? It makes the test easier to read if exactly what was put in the buffer is what we expect our result to be

test/html.test.js
```diff
@@ -1957,3 +2003,65 @@ QUnit.test('removeGopBuffer correctly removes range from buffer', function() {
...
+  try {
+    wrappedSourceBuffer.appendStream();
+  } catch (e) {
+    QUnit.ok(e, 'error rethrown when appendStream fails');
+    QUnit.equal(wrappedSourceBuffer.updating, false,
+      'wrapped source buffer is no longer updating after appendStream fails');
+  }
+
+  try {
+    wrappedSourceBuffer.appendBuffer();
```
ldayananda: should this be `remove`?

src/remove-cues-from-track.js
squarebracket: Yes. If you're using the cue metadata to determine buffer size, then you must not remove cues whose corresponding segments are only being partially removed, otherwise you'd end up with an estimate that undershoots the actual value.

src/virtual-source-buffer.js
squarebracket: I don't think aborts can fail?

src/virtual-source-buffer.js
squarebracket: yeah, that's a good edit for clarity, and the docstring for the function would probably be better as something like
> Returns the equivalent display time of the current GOP's start PTS

src/virtual-source-buffer.js
squarebracket: tbh I just followed what was in `gopsSafeToAlignWith`, but yeah that's a good point.

src/virtual-source-buffer.js
squarebracket: I don't think it's swallowing any other errors, but I might as well gate it for safety as you suggest.

test/html.test.js
squarebracket: It would appear that it should :flushed:

src/remove-cues-from-track.js
ldayananda: I see! I believe this method is also used for textTracks - in that scenario would leaving the overlapping cues cause strange issues?

src/virtual-source-buffer.js
ldayananda: hmm, yeah doesn't look like it's specified at least 👍

src/virtual-source-buffer.js
ldayananda: 💯

src/remove-cues-from-track.js
squarebracket: IIRC, there are two ways this function is called:
1) bounded `(0, currentTime() - backBufferLength)`, where it will remove cues a certain amount of time behind the playhead
2) unbounded `(0, duration())`, when it resets the stream.

In this way, only removing cues that wholly overlap with the bounded removes is probably better, in that you won't accidentally remove a cue that starts before the buffer does but ends after. i.e. if you have a cue (13, 15) and a remove is called for (0, 14), you should be displaying that cue from 14s->15s if you rewind the playhead to 14s.

src/virtual-source-buffer.js
squarebracket: I see that's what's being done now in videojs/http-streaming#153.

test/html.test.js
squarebracket: you mean checking that the event target is implicitly saying the event fired?

test/html.test.js
squarebracket: I take the `ceil` in [the function](https://github.com/videojs/videojs-contrib-media-sources/pull/178/files#diff-7623c1ea12c03a2f94f4e9ae935330f6R104) because of possible rounding errors (and also because that's what `gopsSafeToAlignWith` does it).

test/html.test.js
ldayananda: Yes, exactly

test/html.test.js
ldayananda: Oh, sorry I meant the the ceil used in `pts` on line 1611

squarebracket: Ok wait, I think I was confusing myself, thinking somehow this would important in comparing the numbers. It's not. After thinking about it, I'm pretty sure I put this `ceil`'d here because PTS numbers are actually `int`s, but because of the way JS stores numbers, 2.2*90000=198000.00000000003.

However, I think you've got a good point that dropping the `ceil` would be worth using test data that wouldn't really happen in real life.

src/virtual-source-buffer.js
squarebracket: I really wish I'd written that comment when I wrote the code. It's been so long that I don't remember :sweat:

src/virtual-source-buffer.js
squarebracket: It seems the WHATWG spec [uses the term media timeline](https://html.spec.whatwg.org/#media-timeline) to refer to the thing that's measured in seconds, so I've used that in the function description, and added `(in seconds)` to the `@return`. I'm not sure whether that's more or less ambiguous than `display time`. Let me know what you think.

src/remove-cues-from-track.js
squarebracket: Actually, I thought of a potential problem. Cues in VTT playlists don't necessarily have to end at the duration of the buffer -- they could end after. So a call to remove `(0, duration())` may not catch a cue that hangs over the end of the video.

ldayananda: Did you mean the duration of the buffered region or the whole video's duration? I just double-checked and it looks like WebVTT cues don't necessarily need to end at the video duration so you're right that we could potentially miss a cue hanging over the end of the video and never clear it. Perhaps it would be worth having another utility function to clear all the cues from beginning to end specifically for visible textTracks?

src/virtual-source-buffer.js
ldayananda: This is entirely based on the code, but it seemed like this was walking back one GOP to see if the currentTimePts was before anything in the buffer. I guess the assumption was that the buffer is sorted by pts and we could have overlapping buffer times? I don't know how accurate that is but maybe it jogs your memory

src/virtual-source-buffer.js
ldayananda: I'm in favor of using terminology from relevant specs whenever possible 👍

src/remove-cues-from-track.js
squarebracket: > we could potentially miss a cue hanging over the end of the video and never clear it.

This is what I was talking about.

> Perhaps it would be worth having another utility function to clear all the cues from beginning to end specifically for visible textTracks?

And just do like a `textTrack.cues.forEach((cue) => textTrack.remove(cue))` kind of thing you mean?

ldayananda: yeah, essentially just `removeAllCuesFromTrack(track)` that does `textTrack.cues.forEach((cue) => textTrack.remove(cue))`

src/virtual-source-buffer.js
```diff
@@ -718,16 +769,21 @@ export default class VirtualSourceBuffer extends videojs.EventTarget {
...
+                                             this.mediaSource_.player_.currentTime(),
+                                             this.timeMapping_);
+
+          this.trigger({
+            safeRemovePoint: currGOPStart,
+            segment: tempBuffer,
+            target: destinationBuffer,
+            type: 'bufferMaxed'
           });
+        } else {
```
ldayananda, line 785: Sorry to be pendantic about this but we probably want to restore the old behavior in the else:

```
 if (this.mediaSource_.player_) {
    this.mediaSource_.player_.error({
    code: -3,
    type: 'APPEND_BUFFER_ERR',
    message: error.message,
    originalError: error
  });
}
```

Also, it seems like `throw` is causing the Qunit tests to fail 😂

ldayananda: Oh, forgot to mention the reason we want to do this is because it sets the error state on the tech, http://docs.videojs.com/tech#error, I believe this is also what shows an error dialog on the player

squarebracket: Is that really appropriate, though? This won't be an `APPEND_BUFFER_ERR`; that would be the `QuotaExceededError` that we're catching above.

Not that I disagree with trying to catch uncaught errors for the purpose of notifying the user, generally speaking. But, there could be all sorts of errors thrown at various points in the code. Like the "source buffer already removed" error. So why do you think we should catch these in particular? And what sort of error should we re-throw to the player?

src/remove-cues-from-track.js
squarebracket: I added a `removeAllCuesFromTrack`.

I guess one other option would be to simply pass `Infinity` other than `duration()` for the call to remove all text cues inside contrib-hls, since all cues should fall within [0, Infinity].

src/virtual-source-buffer.js
squarebracket: > I guess the assumption was that the buffer is sorted by pts and we could have overlapping buffer times?

Hmmmmm. But we `break` at the first GOP that has a starting PTS greater than the current time, so I don't know how we'd ever end up in a situation where the previous GOP breaks that condition.

Honestly, this is probably just a programming error on my part. Since it's got the `i === 0` on the front, I wonder if maybe this was just bounds checking to see if any of our buffered GOPs includes the `currentTime`. In which case, I should probably just do that differently. So I'm going to code the change to be that, assuming that's what I was trying to do, because I really don't know what else it would be :\

src/virtual-source-buffer.js
ldayananda: Wouldn't it be an `APPEND_BUFFER_ERR` in the `else` clause though? My main reason for doing this is that we used to actually show all the errors you mention above("source buffer already removed" e.c.t). Otherwise those errors would be lost and the video element may end up in a bad playback with no way to tell that an error occurred

src/virtual-source-buffer.js
ldayananda: Heh, sounds good. I'm in favor of doing things that we remember/understand 😄

src/remove-cues-from-track.js
ldayananda: That would also make sense!

src/virtual-source-buffer.js
```diff
@@ -100,23 +100,23 @@ export const currentGOPStart = (buffer, currentTime, mapping) => {
...
     }
   }
 
-  if (i === 0 || buffer[i - 1].pts > currentTimePts) {
-    return null;
-  }
-
   // This should maybe be Math.floor'd to prevent rounding errors
-  return (buffer[i - 1].pts / 90000) + mapping;
+  return (buffer[gopIndex].pts / 90000) + mapping;
```
ldayananda: Could also simply further by returning from inside the for loop and returning null outside the for loop

src/virtual-source-buffer.js
squarebracket: eeeehhhhhhhhhhhh....... You're right that you used to show errors thrown by `appendBuffer`, but you can also have errors thrown by e.g. `remove`, or `addSourceBuffer`, which are just uncaught exceptions. They were before, they are now.

Is that bad? Yes.

Should they all be caught? Yes.

Should only some of them be caught in this PR, or all of them, or none of them?..... :man_shrugging:

Also, I don't think `APPEND_BUFFER_ERR` / code -3 is defined anywhere, even in videojs-errors.

src/virtual-source-buffer.js
squarebracket: ah yeah good call

forbesjo:
Closing since this repo is going to be archived. I would advise you to open your PR against the next iteration of videojs-contrib-hls at https://github.com/videojs/http-streaming.

